Issue #18, opened 2017-03-13
`InstantDeserializer` is not working with offset of zero `+00:00` and `+00` 


kevinjom:
I am working on a project with Spring boot, and it uses Jackson as the json parse library behind to scene to parse http input message into Objects. And some of our clients will pass a timestamp string like this `2017-03-13T11:11:11.000+00:00`, and I found that didn't work with current implementation of [InstantDeserializer.java](https://github.com/FasterXML/jackson-modules-java8/blob/master/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java#L194). 

I am thinking if we can replace a zone offset of zero that is `+0000` with `Z`, we probably should also replace `+00:00` and `+00` as well, because as I found in this [Wiki](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC) page, they are also valid zero offsets. 

I didn't find a CONTRIBUTION guide so I've made a quick change [here](https://github.com/kevinjom/jackson-modules-java8/commit/183ae5aad731b4460dee6e6e61a15a2eed02cfcc),  I am not sure if you guys are happy with that. If so, I'd like to make a request to fix it.

kevinjom:
Well, I've created a PR https://github.com/FasterXML/jackson-modules-java8/pull/19 here.

cowtowncoder:
Excellent, merged, will be in 2.9.0(-pr2)!

PR #19, opened 2017-03-14
Supported parsing date time text with valid zero zone offsets

kevinjom:
Refer to https://github.com/FasterXML/jackson-modules-java8/issues/18

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -52,7 +52,7 @@ public class InstantDeserializer<T extends Temporal>
     private static final long serialVersionUID = 1L;
 
     /**
-     * Constants used to check if the time offset is zero
+     * Constants used to check if the time offset is zero. See [jackson-modules-java8#18]
      *
      * @since 2.9.0
      */
@@ -65,7 +65,7 @@ public class InstantDeserializer<T extends Temporal>
             a -> Instant.ofEpochMilli(a.value),
             a -> Instant.ofEpochSecond(a.integer, a.fraction),
             null,
-            true // yes, replace +0000 with Z
+            true // yes, replace zero offset with Z
     );
 
     public static final InstantDeserializer<OffsetDateTime> OFFSET_DATE_TIME = new InstantDeserializer<>(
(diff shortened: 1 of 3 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -51,13 +51,21 @@
 {
     private static final long serialVersionUID = 1L;
 
+    /**
+     * Constants used to check if the time offset is zero. See [jackson-modules-java8#18]
+     *
+     * @since 2.9.0
+     */
+    private static final String ISO8601_UTC_ZERO_OFFSET_SUFFIX_REGEX = "\\+00:?(00)?$";
+    private static final String ISO8601_UTC_ZERO_OFFSET_REGEX = "[^\\+]+" + ISO8601_UTC_ZERO_OFFSET_SUFFIX_REGEX;
```
cowtowncoder: This should be pre-compiled as `Pattern` for performance reasons.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -285,6 +288,20 @@ private ZoneId getZone(DeserializationContext context)
         return (_valueClass == Instant.class) ? null : context.getTimeZone().toZoneId();
     }
 
+    private String replaceZeroOffsetAsZIfNecessary(String text)
+    {
+        if (replaceZeroOffsetAsZ && endsWithZeroOffset(text)) {
```
cowtowncoder: This seems to evaluate regexp twice, first to match, then to replace: would it be possible to make that just one operation? Regexp evaluation has non-trivial cost, even considering that textual date handling itself is rather expensive operation.

cowtowncoder:
Thank you for contributing this, yes, makes sense. The only suggestion I have is that since regexp handling has some overhead, it'd be good to do the usual optimization: pre-compile regexp String as `Pattern` (as they are immutable and can be shared, reused), and try to avoid matching and replacing as separate steps.

cowtowncoder:
One other thing: I'm happy to merge this (ideally with changes although I can probably make them myself too if need be), but unless I've asked before we need a filled Contributor License Agreement (CLA) before the first contribution (but just once, it's good for any other patches).
Document is at:

and usually it's easiest to print, fill & sign, scan, email to `info` at fasterxml dot com.
Once this occurs (apologies if I've asked for this before), I can merge the PR.

Thank you again for the contribution!

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
kevinjom: Thanks for the comment @cowtowncoder,  I just checked the doc, your point makes perfect sense, I was wrong, I remember if no match found that would throw an exception. will fix

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
kevinjom: Yes, will fix

kevinjom:
@cowtowncoder I am really grateful for the comments and excited to contribute to this project, I will sign that CLA and send email later (well, before I fixed all the issues you pointed)

kevinjom:
@cowtowncoder comments fixed and CLA mail sent. Please have a look.

cowtowncoder:
@kevinjom Thank you!

PR #22, opened 2017-03-24
#20 Allow `LocalDate` to be serialized/deserialized as number (epoch day)

### LINKED ISSUE #20: Allow `LocalDate` to be serialized/deserialized as number (epoch day)
The `LocalDateSerializer` only supports serializing `LocalDate` instances as strings or arrays, both of which suffer from localization issues ("is it year-month-day or year-day-month?").

An alternative would be to use `LocalDate#toEpochDay()` and `LocalDate#ofEpochDay(long)`, thus allowing a (less human readable but) unambiguous format.

kewne:
Fixes #20


cowtowncoder:
Ok, sounds good to me.
Have I already asked for a CLA? If not, I'd need one from:

and usually easiest way is to print, fill & sign, scan, email to `info` at fasterxml dot com.
Only needs to be done once, good for any number of contributions.

If this could be done, I can merge this in; seemed good to me (I'll add comments if I have any, so far I don't).

cowtowncoder:
Received CLA, merging.

PR #23, opened 2017-03-25
Support single valued and empty arrays for all supported types

### LINKED ISSUE #21: DeserializationFeature.UNWRAP_SINGLE_VALUE_ARRAYS not respected
the following JSON
```
{"zdt":["2017-03-22T14:50:08.121+01:00"]}
```
Cannot be deserialized into the following class:
```
public static class TestJava8TimeBean{
		private ZonedDateTime zdt=ZonedDateTime.now();
		
		public TestJava8TimeBean(){
			
			
		}
		
		public ZonedDateTime getZdt() {
			return zdt;
		}
		
		public void setZdt(ZonedDateTime zdt) {
			this.zdt = zdt;
		}
``` 
But the mapper I use defines:
```
mapper.configure(DeserializationFeature.UNWRAP_SINGLE_VALUE_ARRAYS, true);
```
So using an array with one value should work. It does with strings. Also probably if DeserializationFeature.ACCEPT_EMPTY_ARRAY_AS_NULL_OBJECT is enabled, an empty array should result in a null value.

JPMoresmau:
Should fix #21 , added tests for all types to verify behavior for non empty and empty arrays.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
@@ -76,6 +76,9 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
                 // 20-Apr-2016, tatu: Related to [databind#1208], can try supporting embedded
                 //    values quite easily
                 return (Duration) parser.getEmbeddedObject();
+                
+            case JsonTokenId.ID_START_ARRAY:
+            	return _deserializeFromArray(parser, context);
         }
         return _reportWrongToken(parser, context, JsonToken.VALUE_STRING,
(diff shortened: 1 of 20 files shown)


JPMoresmau:
On my machine I had to change the parent pom.xml to use the latest version of databind with https://github.com/FasterXML/jackson-databind/commit/60ae6000d361f910ab0d7d269a5bac1fc66f4cd9

cowtowncoder:
Excellent. Yes, method for convenient handling of arrays was just added, so requires post 2.9.0.pr2 `jackson-databind`.

I'll be happy to merge this. Just one formality (unless I've already this asked for this): Contributor License Agreement, from: 

usually easiest to print, fill & sign, scan, email to `info` at fasterxml dot com.
Only needs to be done once, good for all contributions for Jackson projects.
If we could get that I'll merge this ASAP.

Thank you in advance!

JPMoresmau:
OK, sent you a CLA via email.

cowtowncoder:
Excellent, merging!

PR #25, opened 2017-04-14
Add base serialization support for Stream

jBouyoud:
This is an implementation proposal for stream serialization support.

### DIFF
--- a/datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/StreamSerializer.java
+++ b/datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/StreamSerializer.java
@@ -1,0 +1,67 @@
+package com.fasterxml.jackson.datatype.jdk8;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.*;
+import com.fasterxml.jackson.databind.ser.ContextualSerializer;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import java.io.IOException;
+import java.util.stream.Stream;
+
+/**
+ * Common typed stream serializer
+ *
+ * @param <T> type of the stream elements
+ */
+public class StreamSerializer<T> extends StdSerializer<Stream<T>> implements ContextualSerializer {
+    /**
+     * Stream elements type (matching T)
+     */
+    private final JavaType elemType;
+    /**
+     * element specific serializer, if any
+     */
+    private transient JsonSerializer<T> elemSerializer = null;
+
+    /**
+     * Constructor
+     *
+     * @param streamType Stream type
+     * @param elemType   Stream elements type (matching T)
+     */
+    public StreamSerializer(JavaType streamType, JavaType elemType) {
+        super(streamType);
+        this.elemType = elemType;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public JsonSerializer<?> createContextual(SerializerProvider provider, BeanProperty property) throws JsonMappingException {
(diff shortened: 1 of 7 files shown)


datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/StreamSerializer.java
```diff
@@ -0,0 +1,67 @@
...
+    public StreamSerializer(JavaType streamType, JavaType elemType) {
+        super(streamType);
+        this.elemType = elemType;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public JsonSerializer<?> createContextual(SerializerProvider provider, BeanProperty property) throws JsonMappingException {
+        if (!elemType.hasRawClass(Object.class) && (provider.isEnabled(MapperFeature.USE_STATIC_TYPING) || elemType.isFinal())) {
+            elemSerializer = (JsonSerializer<T>) provider.findPrimaryPropertySerializer(elemType, property);
```
cowtowncoder: Quick note: modifying instance is not allowed, `createContextual()` MUST create and return a new instance -- otherwise cached "blueprint" instance is modified.
So there needs to be bit more plumbing to create new instance (define abstract factory method here, let sub-classes implement).

datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/WrappedIOException.java
```diff
@@ -0,0 +1,31 @@
...
+    public WrappedIOException(IOException cause) {
+        super(cause);
+    }
+
+    /**
+     * Returns the wrapped {@link IOException}
+     *
+     * @return the wrapped {@link IOException}
+     */
+    @Override
```
cowtowncoder: Why override (and especially as `synchronized`)?

cowtowncoder:
Thank you -- this looks like a good starting point.
There are a few potential extensions that ideally would be handled (since it's serialized as an array, perhaps "unwrap if single item"; `JsonInclude` filtering for empty -- if possible, not sure if it is?), but this is probably good enough as is.
The only thing that needs to be changed is the `createContextual` thing I added comment for.

At this point I'd be happy to merge this (I can make change if need be), but unless I have already requested it, we'd need Contributor License Agreement:

before first contribution (if I have asked & received one, apologies -- so many PRs these days!)
Easiest way is usually to print, fill & sign, scan, email to `info` at fasterxml dot com.
Once this happens, I'll be happy to merge the PR and it will get in 2.9.0.pr3, due out relatively soon.

Thank you once again for contributing this -- it has been requested by a few. And I especially like the fact that it can expose exceptions appropriately, nice work.

datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/WrappedIOException.java
jBouyoud: override to handle IOException cast. synchronised is a mistake.

jBouyoud:
Ok CLA sent by email

cowtowncoder:
Excellent, thank you!

datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/WrappedIOException.java
jmax01: You need to close those streams in a finally.
Otherwise if they can hold resources.

jmax01: Actually should be handled via try-with-resources.

jmax01: Also shouldn't Stream serializers extend `AsArraySerializerBase`?

Issue #27, opened 2017-05-03
Stream Serializers need to close.


jmax01:
The Stream Serializers do not close on error this can lead to leaks if the value supplier has open resources (streaming from a ResultSet for example).

I have a PR ready,

cowtowncoder:
Sounds good to me, let me check.

cowtowncoder:
I think this was added as per merged PR, so closing.

PR #28, opened 2017-05-03
Stream Serializers need to close. #27

jmax01:
Jdk8Serializers: Should not be parameterized.
StreamDoubleSerializer: Always close stream.  
StreamIntSerializer: Always close stream.
StreamLongSerializer: Always close stream.
StreamSerializer: Always close stream.


cowtowncoder:
Makes sense, will merge. Thank you!

Issue #29, opened 2017-05-08
Rename primitive Stream serializers to match their actual associated Stream implementations


jmax01:
Currently primitive java.util.stream.* serializers are named Stream[Primitive]Serializer which means they do not match the names in of their actual stream implementations.

I suspect this an attempt to group all the serializers together when viewing the package contents.

I would proffer that this violates the principle of least surprise.

PR #30, opened 2017-05-08
#29 Rename primitive Stream serializers to match their actual associated Stream implementations

jmax01:
Stream implementations.

DoubleStreamSerializer: Renamed from StreamDoubleSerializer.
IntStreamSerializer: Renamed from StreamIntSerializer.
LongStreamSerializer: Renamed from StreamLongSerializer.

StreamTest: Deleted and replaced with Serializer specific tests.
StreamCloseTest: Deleted and replaced with Serializer specific tests.

ModuleTestBase: Made mapperWithModule and mapperWithModule(boolean) 
static to allow use in test class that are not subclasses.
This allows use of more modern junit constructs like @Rule.


cowtowncoder:
Thanks!

Issue #65, opened 2018-04-03
Use `DeserializationContext.handleWeirdXxxValue()` for datetime deserializers


remal:
Lets look on LocalDateTimeDeserializer (for example):
```java
public LocalDateTime deserialize(JsonParser parser, DeserializationContext context) throws IOException {
    if (parser.hasTokenId(JsonTokenId.ID_STRING)) {
        String string = parser.getText().trim();
        // ...
        try {
            // ...
            return LocalDateTime.parse(string, _formatter);
        } catch (DateTimeException e) {
            _rethrowDateTimeException(parser, context, e, string);
        }
    }
    // ...
}
```

Please change it to:
```java
public LocalDateTime deserialize(JsonParser parser, DeserializationContext context) throws IOException {
    if (parser.hasTokenId(JsonTokenId.ID_STRING)) {
        String string = parser.getText().trim();
        // ...
        try {
            // ...
            return LocalDateTime.parse(string, _formatter);
        } catch (DateTimeException e) {
            try {
                return context.handleWeirdStringValue(LocalDateTime.class, string, e.getMessage());
            } catch (Throwable t) {
                t.initCause(e);
            }
        }
    }
    // ...
}
```

Why do I need this? I want to create a module that defines some DeserializationProblemHandlers that allow to deserialize datetime types using several formats. Now I need to create mix-ins to override datetime deserializers.

Please apply this change to all deserializers and key deserializers.

cowtowncoder:
Makes sense. I'll add notes to PR itself.

PR #66, opened 2018-04-03
Using DeserializationContext.handleWeirdStringValue for DateTimeException

remal:
This is an example implementation for issue #65


remal:
@cowtowncoder 
I've added some unit tests for LocalDateTIme deserialization. Please take a look.
Also I've sent signed CLA to the provided email.

cowtowncoder:
Looks good! Will make one small change wrt key deserializer -- needs to call `handleWeirdKey()` instead (since key handling differs a bit from value handling), but that's the only thing I'll change.

PR #75, opened 2018-05-27
Use `SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS` for configuring `Duration` serialization

kezhuw:
The current implementation uses `WRITE_DATES_AS_TIMESTAMPS` and ignores `WRITE_DURATIONS_AS_TIMESTAMPS` totally.

I thought it would be less surprising to use `WRITE_DURATIONS_AS_TIMESTAMPS` for time duration serialization.


cowtowncoder:
Thank you for submitting this PR! Since this changes observed behavior, it needed to go in a new minor version, and now that I branched `2.10` I was able to merge the suggested change in manually.

One other thing that might be worth considering is whether handling of `Period` should also allow timestamp notation (and if so, feature to use should have same as with `Duration`. But since timestamps are not yet supported that's separate issue.

Thank you again for the improvement; this will be in 2.10.0 once that gets released.

Issue #78, opened 2018-09-09
Year deserialization ignores `@JsonFormat` pattern


devdevx:
Jackson version: 2.9.3
Jackson datatype version: 2.9.6

Example of the `@JsonFormat` definition:
```java
public class ObjectTest {
   @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "'Y'yyyy")
    private Year customYear;
   // Getters and setters
}
```

Test example:
```java
ObjectTest object = new ObjectTest();
object.setCustomYear(Year.of(2018));
ObjectMapper objectMapper = new ObjectMapper().registerModule(new JavaTimeModule());
String json = objectMapper.writeValueAsString(object); // OK -> "{\"customYear\":\"Y2018\"}";
object = objectMapper.readValue(json, ObjectTest.class); // ERROR -> Throws exception
```

**Possibly related to**: https://github.com/FasterXML/jackson-modules-java8/issues/51

Stack trace:
```java
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.Year` from String "Y2018": Failed to deserialize java.time.Year: (java.time.format.DateTimeParseException) Text 'Y2018' could not be parsed at index 0
 at [Source: (String)"{"customYear":"Y2018"}"; line: 1, column: 15] (through reference chain: com.devdevx.examples.jackson.domain.javatime.JavaTime["customYear"])

	at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:67)
	at com.fasterxml.jackson.databind.DeserializationContext.weirdStringException(DeserializationContext.java:1548)
	at com.fasterxml.jackson.databind.DeserializationContext.handleWeirdStringValue(DeserializationContext.java:910)
	at com.fasterxml.jackson.datatype.jsr310.deser.JSR310DeserializerBase._handleDateTimeException(JSR310DeserializerBase.java:80)
	at com.fasterxml.jackson.datatype.jsr310.deser.YearDeserializer.deserialize(YearDeserializer.java:64)
	at com.fasterxml.jackson.datatype.jsr310.deser.YearDeserializer.deserialize(YearDeserializer.java:34)
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:127)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:287)
[...]

PR #79, opened 2018-09-14
Solve #78, Year deser use custom pattern

devdevx:
I'm not sure in the definition of the `DEFFAULT_PATTERN`, I have used the same code that is used to detine the year part in the `java.time.format.DateTimeFormater.ISO_LOCAL_DATE`


cowtowncoder:
Actually, the original implementation handled it the way I think works ok: `null` is used to mark "just use default", which just calls:

    Year.parse(string);

so that way I think we can avoid having to use a default.

cowtowncoder:
Not sure why the merge conflict; patched manually.

cowtowncoder:
Manually merged: thank you! This will be in 2.9.7 release

PR #84, opened 2018-10-02
code to handle FasterXML/jackson-databind/issues/2141

abracadv8:
Guards against numbers causing CPU or OOM issues when deserializing
large numbers into Instant or Duration (see https://github.com/FasterXML/jackson-databind/issues/2141
)  by either:

* Scientific notation too large (eg 10000e100000)
* Raw string repesenting a number of length too long

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
@@ -16,6 +16,7 @@
 
 package com.fasterxml.jackson.datatype.jsr310.deser;
 
+import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.JsonTokenId;
@@ -40,6 +41,11 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
 
     public static final DurationDeserializer INSTANCE = new DurationDeserializer();
 
+    private static final BigDecimal INSTANT_MAX = new BigDecimal(
+            java.time.Instant.MAX.getEpochSecond() + "." + java.time.Instant.MAX.getNano());
+    private static final BigDecimal INSTANT_MIN = new BigDecimal(
+            java.time.Instant.MIN.getEpochSecond() + "." + java.time.Instant.MIN.getNano());
+
     private DurationDeserializer()
     {
         super(Duration.class);
@@ -52,6 +58,13 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
         {
             case JsonTokenId.ID_NUMBER_FLOAT:
                 BigDecimal value = parser.getDecimalValue();
+                // If the decimal isnt within the bounds of a float, bail out
+                if(value.compareTo(INSTANT_MAX) > 0 ||
+                        value.compareTo(INSTANT_MIN) < 0) {
+                    throw new JsonParseException(context.getParser(),
+                            "Value of Float too large to be converted to Duration");
+                }
+
                 long seconds = value.longValue();
                 int nanoseconds = DecimalUtils.extractNanosecondDecimal(value, seconds);
                 return Duration.ofSeconds(seconds, nanoseconds);
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
@@ -16,7 +16,6 @@
 
 package com.fasterxml.jackson.datatype.jsr310.deser;
 
-import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.JsonTokenId;
@@ -41,10 +40,10 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
 
     public static final DurationDeserializer INSTANCE = new DurationDeserializer();
 
-    private static final BigDecimal INSTANT_MAX = new BigDecimal(
-            java.time.Instant.MAX.getEpochSecond() + "." + java.time.Instant.MAX.getNano());
-    private static final BigDecimal INSTANT_MIN = new BigDecimal(
-            java.time.Instant.MIN.getEpochSecond() + "." + java.time.Instant.MIN.getNano());
+    private static final BigDecimal DURATION_MAX = new BigDecimal(
+            Long.MAX_VALUE + "." + java.time.Instant.MAX.getNano());
+    private static final BigDecimal DURATION_MIN = new BigDecimal(
+            Long.MIN_VALUE + "." + java.time.Instant.MIN.getNano());
 
     private DurationDeserializer()
     {
@@ -58,11 +57,11 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
         {
             case JsonTokenId.ID_NUMBER_FLOAT:
                 BigDecimal value = parser.getDecimalValue();
-                // If the decimal isnt within the bounds of a float, bail out
-                if(value.compareTo(INSTANT_MAX) > 0 ||
-                        value.compareTo(INSTANT_MIN) < 0) {
-                    throw new JsonParseException(context.getParser(),
-                            "Value of Float too large to be converted to Duration");
+                // If the decimal isn't within the bounds of a duration, bail out
+                if(value.compareTo(DURATION_MAX) > 0 ||
+                        value.compareTo(DURATION_MIN) < 0) {
+                    throw new DateTimeException(
+                            "Instant exceeds minimum or maximum Duration");
                 }
 
                 long seconds = value.longValue();
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -17,6 +17,7 @@
 package com.fasterxml.jackson.datatype.jsr310.deser;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.JsonTokenId;
@@ -29,6 +30,8 @@ import com.fasterxml.jackson.datatype.jsr310.DecimalUtils;
 
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
 import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.OffsetDateTime;
@@ -51,6 +54,11 @@ public class InstantDeserializer<T extends Temporal>
     extends JSR310DateTimeDeserializerBase<T>
 {
     private static final long serialVersionUID = 1L;
+
+    private static final BigDecimal INSTANT_MAX = new BigDecimal(
+            java.time.Instant.MAX.getEpochSecond() + "." + java.time.Instant.MAX.getNano());
+    private static final BigDecimal INSTANT_MIN = new BigDecimal(
+            java.time.Instant.MIN.getEpochSecond() + "." + java.time.Instant.MIN.getNano());
 
     /**
      * Constants used to check if the time offset is zero. See [jackson-modules-java8#18]
@@ -277,8 +285,15 @@ public class InstantDeserializer<T extends Temporal>
                 timestamp, this.getZone(context)));
     }
     
-    protected T _fromDecimal(DeserializationContext context, BigDecimal value)
+    protected T _fromDecimal(DeserializationContext context, BigDecimal value) throws JsonParseException
     {
+        // If the decimal isnt within the bounds of an Instant, bail out
+        if(value.compareTo(INSTANT_MAX) > 0 ||
+                value.compareTo(INSTANT_MIN) < 0) {
+            throw new JsonParseException(context.getParser(),
+                    "Value of String too large to be converted to Instant");
+        }
+
         long seconds = value.longValue();
         int nanoseconds = DecimalUtils.extractNanosecondDecimal(value, seconds);
         return fromNanoseconds.apply(new FromDecimalArguments(
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -285,13 +285,13 @@ public class InstantDeserializer<T extends Temporal>
                 timestamp, this.getZone(context)));
     }
     
-    protected T _fromDecimal(DeserializationContext context, BigDecimal value) throws JsonParseException
+    protected T _fromDecimal(DeserializationContext context, BigDecimal value)
     {
         // If the decimal isnt within the bounds of an Instant, bail out
         if(value.compareTo(INSTANT_MAX) > 0 ||
                 value.compareTo(INSTANT_MIN) < 0) {
-            throw new JsonParseException(context.getParser(),
-                    "Value of String too large to be converted to Instant");
+            throw new DateTimeException(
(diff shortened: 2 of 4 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -48,22 +58,35 @@ private DurationDeserializer()
...
         {
             case JsonTokenId.ID_NUMBER_FLOAT:
                 BigDecimal value = parser.getDecimalValue();
+                // If the decimal isnt within the bounds of a float, bail out
+                if(value.compareTo(INSTANT_MAX) > 0 ||
+                        value.compareTo(INSTANT_MIN) < 0) {
+                    NumberFormat formatter = new DecimalFormat("0.0E0");
+                    throw new JsonParseException(context.getParser(),
+                            String.format("Value of BigDecimal (%s) not within range to be converted to Duration",
+                                    formatter.format(value)));
```
toddjonker: Could this use `string` rather than reformatting the parsed value?

toddjonker:
Is there a potential for this same problem -- `BigDecimal.longValue()` taking unbounded time -- to affect other conversions?  For example, what if the serialized decimal is being mapped to an `Integer` or `long` instead of an `Instant` or `Duration`?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -48,22 +58,35 @@ private DurationDeserializer()
     @Override
     public Duration deserialize(JsonParser parser, DeserializationContext context) throws IOException
     {
+        String string = parser.getText().trim();
```
yishaigalatzer: this will allocate twice as much than before. I would keep the string inside the switch statement, or alternatively see if there is a way to get the length from the parser without allocating a new string

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/TestInstantSerialization.java
```diff
@@ -376,6 +377,84 @@ public void testDeserializationWithTypeInfo04() throws Exception
...
+     * This test can potentially hang the VM, so exit if it doesn't finish
+     * within a few seconds.
+     *
+     * @throws Exception
+     */
+    @Test(timeout=3000, expected = JsonParseException.class)
+    public void testDeserializationWithTypeInfoAndStringTooLarge03() throws Exception
+    {
+        // Build a big string by hand
+        String number = "0000000000000000000000000";
```
yishaigalatzer: would using an actual number rather than 0 be more resilient to future code changes?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
abracadv8: If the string is a 1,000,000 characters long, I would assume printing it might be more trouble than what its worth, --  eg the test case where I append "1000000000" a thousand times would end up with a 1mb long exception message.

I can just not print the string and just say its not within the expected range.

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/TestInstantSerialization.java
abracadv8: I can, but how many zeroes would be reasonable 100?
I believe the current java Instant has a max length of ~30ish digits.

abracadv8:
@toddjonker  -- Yes, I would assume the same thing would happen for other areas.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
abracadv8: I can try getTextLength() -- I don't know what the implementation of that is however (eg, if it allocates memory or not).

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
cowtowncoder: I think it'd be better to just include length of String as presumably this should only be hit for attack, or some other use case in which input is not legitimate. So printing that value does not make sense.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -48,22 +58,35 @@ private DurationDeserializer()
...
+                            String.format("Value of BigDecimal (%s) not within range to be converted to Duration",
+                                    formatter.format(value)));
+                }
+
                 long seconds = value.longValue();
                 int nanoseconds = DecimalUtils.extractNanosecondDecimal(value, seconds);
                 return Duration.ofSeconds(seconds, nanoseconds);
 
             case JsonTokenId.ID_NUMBER_INT:
+                if(string.length() > INSTANT_MAX_STRING_LEN) {
```
cowtowncoder: Is the attack also possible via `BigInteger`, or just `BigDecimal`? I am thinking that (as per earlier comments by others) for common case there would be unnecessary allocation of another String, for the common case of input being relatively short.
And for special case of binary formats there would also be conversion from binary number into String.

So the extra checks would mostly make sense if use of `BigInteger` is also similarly dangerous.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
abracadv8: Makes sense.

abracadv8:
QQ @cowtowncoder  - Is there a specific form / CLA to fill out for this contribution?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
toddjonker: The DoS is not coupled to the length of the string, in fact the problem is that short input strings can cause the failure, by representing very large numbers.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
toddjonker: It's just `BigDecimal`, because `BigDecimal.longValue()` (and `intValue()` and `scale()` and some others) will consume unbounded resources even given short string representations of the decimal value.  It's the conversion to an integer that's the problem.

The input-string length checks in this PR are unrelated to [the reported DoS vector](https://github.com/FasterXML/jackson-databind/issues/2141).  IMO they should be submitted and considered separately, because they are muddying the analysis here by conflating independent concerns.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
abracadv8: I've removed this code entirely.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
abracadv8: I have removed all the string length checking entirely and it only contains the code to handle exponential numbers.

raganhan:
Please take a look at my comment on https://github.com/FasterXML/jackson-databind/issues/2141#issuecomment-426833951

Bound checking is necessary but not sufficient as the problem also happens for large **negative** exponents, e.g. `new BigDecimal("1e-100000000")`

cowtowncoder:
@abracadv8 Basic Jackson CLA from https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf works, if you haven't sent one (and are not covered by one of existing Corporate CLAs we have received)

arteam:
There is a test for `1e-100000000000`, but there is no test for `1e-1000000000`, which is high enough not to be considered as zero.

arteam:
I think it should posible to guard against small numbers by checking the scale. `_fromDecimal` will be invoked only in case the client passes a decimal value where the fraction part represent nanoseconds (like `1538326357.298509112`). Only the first 9 digits are used for that, so, practically, we can reject all numbers with scale more than 9.

abracadv8:
> @abracadv8 Basic Jackson CLA from https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf works, if you haven't sent one (and are not covered by one of existing Corporate CLAs we have received)

Sent.  

I'm contributing this under my personal time/resources, but I'm going to submit it to my company and fill out a second if they approve.

yishaigalatzer:
@cowtowncoder just my take:

I would write a test that validates the behavior of inputs just above "too big", and just below "too small" prior to the change. Run these tests with the existing code prior to the change, and then make the fix to keep the same behavior, thus preventing unexpected behavior changes, regardless what is the more correct behavior you come up with in a clean design.

Why? This library is already super popular, and it is hard to tell how existing code depends on that edge case. Making a change in behavior in a minor version is not justified.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -277,8 +285,15 @@ protected T _fromLong(DeserializationContext context, long timestamp)
                 timestamp, this.getZone(context)));
     }
     
-    protected T _fromDecimal(DeserializationContext context, BigDecimal value)
+    protected T _fromDecimal(DeserializationContext context, BigDecimal value) throws JsonParseException
     {
+        // If the decimal isnt within the bounds of an Instant, bail out
+        if(value.compareTo(INSTANT_MAX) > 0 ||
+                value.compareTo(INSTANT_MIN) < 0) {
+            throw new JsonParseException(context.getParser(),
```
toddjonker: Today, this method throws `DateTimeException` when the given value is out of bounds, so we should throw that here, else this is a backwards-incompatible change.

This test case:
```
    @Test // (expected = DateTimeException.class)                 // Disabled to provoke failure
    public void testDeserializationAsFloat04() throws Exception
    {
        Instant date = Instant.MAX;
        String customInstant = (date.getEpochSecond() + 1) + ".0";
        MAPPER.readValue(customInstant, Instant.class);
    }
```
results in:
```
java.time.DateTimeException: Instant exceeds minimum or maximum instant

	at java.time.Instant.create(Instant.java:411)
	at java.time.Instant.ofEpochSecond(Instant.java:330)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.lambda$static$1(InstantDeserializer.java:66)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer._fromDecimal(InstantDeserializer.java:284)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.deserialize(InstantDeserializer.java:171)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.deserialize(InstantDeserializer.java:50)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4013)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3004)
	at com.fasterxml.jackson.datatype.jsr310.TestInstantSerialization.testDeserializationAsFloat04(TestInstantSerialization.java:244)
```

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -52,6 +58,13 @@ public Duration deserialize(JsonParser parser, DeserializationContext context) t
         {
             case JsonTokenId.ID_NUMBER_FLOAT:
                 BigDecimal value = parser.getDecimalValue();
+                // If the decimal isnt within the bounds of a float, bail out
+                if(value.compareTo(INSTANT_MAX) > 0 ||
+                        value.compareTo(INSTANT_MIN) < 0) {
+                    throw new JsonParseException(context.getParser(),
+                            "Value of Float too large to be converted to Duration");
```
toddjonker: Turns out this isn't the correct boundary: `Duration` can hold a number-of-seconds up to `Long.MAX_VALUE`, which is larger than `Instant.MAX.getEpochSecond()`.

For larger values, Jackson currently has the troubling behavior of `BigDecimal.longValue()`.  This test passes:
```
    @Test
    public void testDeserializationAsFloat05() throws Exception
    {
        String customInstant = new BigInteger(Long.toString(Long.MAX_VALUE)).add(BigInteger.ONE) + ".0";
        Duration value = READER.without(DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS)
                                 .readValue(customInstant);
        assertEquals(Long.MIN_VALUE, value.getSeconds());   // The Duration is negative!
    }
```

toddjonker: For reference on behavior of similar overflow, `Duration.of(Long.MAX_VALUE, ChronoUnit.MINUTES)` throws:
```
java.lang.ArithmeticException: long overflow
```

while `Duration.parse("PT" + Long.MAX_VALUE + "0S")` -- note the added `0` to force overflow -- throws:
```
java.time.format.DateTimeParseException: Text cannot be parsed to a Duration: seconds
```

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -52,6 +58,13 @@ public Duration deserialize(JsonParser parser, DeserializationContext context) t
         {
             case JsonTokenId.ID_NUMBER_FLOAT:
                 BigDecimal value = parser.getDecimalValue();
+                // If the decimal isnt within the bounds of a float, bail out
```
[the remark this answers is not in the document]
abracadv8: Changed to `// If the decimal isn't within the bounds of a duration, bail out` since a duration has different bounds than an instant.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
abracadv8: I swapped the checks to throw `throw new DateTimeException("Instant exceeds minimum or maximum instant");`
and `throw new DateTimeException("Instant exceeds minimum or maximum Duration");` respectively.

I updated the bounds for duration to be the following, although I am not 100% certain it is correct with regards to minimum: 
```
    private static final BigDecimal DURATION_MAX = new BigDecimal(
            Long.MAX_VALUE + "." + java.time.Instant.MAX.getNano());
    private static final BigDecimal DURATION_MIN = new BigDecimal(
            Long.MIN_VALUE + "." + java.time.Instant.MIN.getNano());
```

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/TestDurationDeserialization.java
```diff
@@ -53,28 +54,36 @@ public void testDeserializationAsFloat03() throws Exception
         assertEquals("The value is not correct.", Duration.ofSeconds(13498L, 8374), value);
     }
 
+    @Test
+    public void testDeserializationAsFloat04() throws Exception
```
toddjonker: This looks the same as the previous test case.  Am I missing something?

toddjonker:
I'm just submitted PR #85 adding a *bunch* of test cases around the boundary edges.  I believe it's a superset of what's tested here.

I suggest we get that reviewed and pulled and then this change can layered on top of it.  The fix in that case wouldn't need any test changes, expect to add some low timeouts to the (8) currently-slow test cases.

cowtowncoder:
Ok, I merged @toddjonker's alternative, which I think truncates all out-of-bounds cases to zero value (that is, won't bother with upper-bound check for truncation or exception), which seems like 
one way to deal with that.
Also created actual issue for this repo, #90, to use for release notes.

Will close this PR but we can continue discussion either on #90 comments, new issue, mailing list or gitter. I think there are still remaining potential issues that could be tackled for 2.9.8.

1Jesper1:
Is there already a CVE issued for this problem? When will 2.9.8 be released?

cowtowncoder:
@1Jesper1 I haven't requested CVE, but if someone has or is aware of one, I can add that.
As to 2.9.8, as soon as I have time to work on couple of serialization gadgets so they can be included.
Lately I haven't had much time to work on Jackson (does not pay the bills), so this may take a while.
But hopefully before end of November.

PR #87, opened 2018-10-18
Prevent unbounded latency converting decimals to time

toddjonker:
This change prevents latency explosions when working with high-magnitude `BigDecimal` values, while preserving the current behavior on the edge cases.

It's fixes the original bug reported in https://github.com/FasterXML/jackson-databind/issues/2141 but not the several related issues discussed therein. It builds atop my earlier PR https://github.com/FasterXML/jackson-modules-java8/pull/85

One challenge here is that the current two-step conversion process, using `BigDecimal.longValue()` and `DecimalUtils. extractNanosecondDecimal()` separately, made it hard to control the edge cases because information is lost during the former.  So I combined the two into a single helper method.

(IMO these helpers shouldn't be public API, since they are bespoke semantics needed by this package, and unlikely to be usable by other contexts.  Frankly I'd prefer to make the new helper method package-protected, if that's okay with the maintainers.)

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/DecimalUtils.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/DecimalUtils.java
@@ -17,6 +17,7 @@
 package com.fasterxml.jackson.datatype.jsr310;
 
 import java.math.BigDecimal;
+import java.util.function.BiFunction;
 
 /**
  * Utilities to aid in the translation of decimal types to/from multiple parts.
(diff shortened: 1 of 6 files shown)


cowtowncoder:
Ok, yes, excellent thank you for the fix!

+1 for hiding functionality (I often use `protected` as compromise to allow sub-classing but not advertise it). I also think that deprecation here makes sense, and then dropping method off in 3.0.

I can make that change after merge.

Let's see if I can continue discussion on remaining challenge after I merge this.

PR #104, opened 2019-02-27
output LocalTime json schema format as Time

jaisonpjohn:
Similar to and modeled after https://github.com/FasterXML/jackson-datatype-jsr310/pull/78

When I am generating Swagger spec for a LocalTime Java Object I wanted to get 
`"format" : "time"` instead of `"format" : "date-time"`.
Otherwise it was tripping off a codegen-from-swagger on consumer side.

Tested locally. Do you need a test for this? If so, can you point me to an example existing test?

Thanks for your time.


cowtowncoder:
Ended up manually merging so closing PR -- thank you again for submitting this!

PR #110, opened 2019-04-09
#109 Add Instant deserialization with nanoseconds in case with Long type 

### LINKED ISSUE #109: Probably bug with Instant deserialization
When i deserialize Instant and use DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS i expect that value - for example (946684800900000000) will deserialize successfully, but only values with point like these (946684800.900000000) work correct. I fixed it and can create pull request

AntonKuzValid:
(to fix #109)

I have added correct deserialization of java.time.Instant class in case when you pass value without dot


cowtowncoder:
Closed #109, will close PR as well. Does not mean there is no remaining issue, just that I don't think this is a right approach.

Issue #58, opened 2018-03-07
Should not parse `LocalDate`s from number (timestamp) -- or at least should have an option preventing


billoneil:
In Jackson 2.8.x I had a unit test that parsed the following and expected an exception.

```json
{
  "date": 123,
  "message": "Happy New Year!"
}
```

As of 2.9 related to this commit https://github.com/FasterXML/jackson-modules-java8/pull/22/files#diff-a1552641fab3c9690f2c8b0f5ee2fe89R102 it now parses as `1970-05-04`

Is there a way to turn this off if I expect my dates to come in as a different format? I don't want to parse random integers as dates when it should be invalid input.

cowtowncoder:
No mechanism to prevent that currently; integers are accepted by default for all date/time types (so change for 2.9, if it didn't work on 2.8) is more due to unifying handling than otherwise intended change.

But come to think of it now, it is not actually clear that numbers should work for `LocalDate`, given that it is pure date value with no canonical numeric value without timezone.
So perhaps it is something that should be blocked in 3.0. Since 2.9.x are already out, behavior can not be changed in a patch however.

As to blocking value you will need to register a custom deserializer that does this.

billoneil:
Just thought I would mention the change in behavior to see if it was intended or not. I can work around it on my end if needed. Feel free to close if you don't plan on changing the behavior.

cowtowncoder:
@billoneil Looks like this was due to #20 actually, so it was intentional. Although perhaps bit misguided, considering the part.

However! I do have one possibly useful idea wrt 2.9.x, your specific use case: there is property `lenient` for `@JsonFormat`; and that defaults to `true`, but may be configured differently at 3 levels: global default, per-type (`LocalDate`) and per-property.
I don't think this is yet supported by this particular deserializer, but it could be; and I think it would make sense... especially if coupled with format `Shape` for deserialization (not just serialization).

It is some work but it seems to me like a good way to do this -- my intent going forward is to limit number of on/off feature flags that are type-specific, and `@JsonFormat` (along with Config Overrides that allow specifying overrides for types and global default(s)).

cowtowncoder:
I think that there should be a way to prevent this, but I am  not sure what the correct path would be.
A new module-configuration feature is a possibility, I suppose.

kupci:
> _I think that there should be a way to prevent this, but I am not sure what the correct path would be.
> A new module-configuration feature is a possibility, I suppose._
> 

[...]

PR #139, opened 2019-09-24
issue 58 - NUMBER_INT should be specified when deserializing LocalDate as EpochDays

kupci:
fix for issue #58 NUMBER_INT should be specified when deserializing LocalDate as EpochDays (unless lenient)

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
@@ -7,6 +7,7 @@ import java.util.Locale;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonFormat.Feature;
+import com.fasterxml.jackson.annotation.JsonFormat.Shape;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.BeanProperty;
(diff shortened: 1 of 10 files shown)


kupci:
Needs as 'strict' test.

cowtowncoder:
Looks good -- I'll add some comments on adding `@since` tags on new methods.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
```diff
@@ -34,28 +34,46 @@
      */
     protected final boolean _isLenient;
 
+    /**
+     * Flag that indicates the {@link JsonFormat.Shape} annotation.
+     *
+     */
```
cowtowncoder: Add `@since 2.10`

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
```diff
@@ -92,6 +110,12 @@ protected JSR310DateTimeDeserializerBase(JSR310DateTimeDeserializerBase<T> base,
                     deser = deser.withLeniency(leniency);
                 }
             }
+            //if (format.hasShape()) {
+                JsonFormat.Shape shape = format.getShape();
```
cowtowncoder: Probably want to check existing `_shape` just because `createContextual()` might get called multiple times (should not be, but should be define idempotent just in case).

cowtowncoder:
Quick question: this is against `master`: I assume this is because change might not make it in time for `2.10`? If so that's fine; I'll create `2.11` branch right after 2.10.0 release, and probably want to rebase against that. I can of course try cherry-picking to backport, even if merged against master. But I assume we'd like this for 2.x series, if possible?

kupci:
Cool, thanks for review. I'll make requested changes tonight. I have been going to upstream/master more out of habit, but I can certainly create PR against 2.10 if tonight CST is ok. One thing I need to figure out, to improve test, is how to send two config overrides, quick Google search and JavaDoc didn't reveal the technique, so I was thinking of creating a small test class and adding annotation there, which I've seen in other tests.

cowtowncoder:
Ok, I think I did config overrides in a test yesterday. Method is different between 2.10 and 3.0, due to rewrite, but 2.10 does have builder-based approach now too I think.
I think change was in .... LocalDateDeserializerTest? Should be possible to chain.

kupci:
I'll close this and create a PR for 2.10 instead.

cowtowncoder:
As per my email comments, rebase for `2.10` branch is fine; just need to make sure that all existing constructors/methods in base class still exist in patch (deprecated).
And when merging forward (master / 3.0), remove such deprecated methods, only needed within 2.x.

kupci:
@cowtowncoder Ok if I keep this one as-is, since it's ready, and then I will backport to 2.10?

cowtowncoder:
Quick note: I just created `2.11` branch, as that will be needed here. It could have gone in 2.10.0, since minor versions can have some changes in behavior, but I try not to do such changes in patch versions. Problem being that there is small (but existing) chance of breakage.
I will try to keep 2.11 dev cycle much shorter than 2.10, 2.11 pre-release candidates maybe going out by December.

So, merge to `master`, cherry-pick to 2.11; or base off 2.11, either work.

PR #140, opened 2019-10-03
Fixes 127 - ZonedDateTime in map keys can now be serialized with zone IDs

### LINKED ISSUE #127: ZonedDateTime in map keys ignores option to write Zone IDs
As of v2.9.9, the following fails:

```java
class Sample {
    private Map<ZonedDateTime, ZonedDateTime> map;
    private ZonedDateTime dateTime;

    public Sample() {
    }

    public ZonedDateTime getDateTime() {
        return dateTime;
    }

    public void setDateTime(ZonedDateTime dateTime) {
        this.dateTime = dateTime;
    }

    public Map<ZonedDateTime, ZonedDateTime> getMap() {
        return map;
    }

    public void setMap(Map<ZonedDateTime, ZonedDateTime> map) {
        this.map = map;
    }
}

public class JacksonTest {

    @Test
    public void shouldSerializeZoneIdInMapKeys() throws IOException {
        // given
        final ObjectMapper objectMapper = new ObjectMapper();
        objectMapper.registerModule(new Jdk8Module());
        objectMapper.registerModule(new JavaTimeModule());
        objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
        objectMapper.enable(SerializationFeature.WRITE_DATES_WITH_ZONE_ID);

        final ZonedDateTime datetime = ZonedDateTime.parse("2007-12-03T10:15:30+01:00[Europe/Warsaw]");
        final HashMap<ZonedDateTime, ZonedDateTime> map = new HashMap<>();
        map.put(datetime, datetime);
        final Sample sample = new Sample();
        sample.setMap(map);
        sample.setDateTime(datetime);

        // when
        final String value = objectMapper.writeValueAsString(sample);

        // then
        assertThat(value).isEqualTo("{\"map\":" +
                "{\"2007-12-03T10:15:30+01:00[Europe/Warsaw]\":\"2007-12-03T10:15:30+01:00[Europe/Warsaw]\"}," +
                "\"dateTime\":\"2007-12-03T10:15:30+01:00[Europe/Warsaw]\"" +
                "}");
    }
}
```

I'd expect that Zone ID is written not only in map keys but also in map values. 
Currently `SerializationFeature.WRITE_DATES_WITH_ZONE_ID` has no impact on map keys.

Sample project:
[test.zip](https://github.com/FasterXML/jackson-modules-java8/files/3491978/test.zip)

vetler:
Fixes #127 - ZonedDateTime in map keys can now be serialized with zone IDs

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/key/ZonedDateTimeKeySerializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/key/ZonedDateTimeKeySerializer.java
@@ -7,6 +7,7 @@ import java.time.format.DateTimeFormatter;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.SerializerProvider;
 
 public class ZonedDateTimeKeySerializer extends JsonSerializer<ZonedDateTime> {
@@ -21,11 +22,14 @@ public class ZonedDateTimeKeySerializer extends JsonSerializer<ZonedDateTime> {
     public void serialize(ZonedDateTime value, JsonGenerator gen, SerializerProvider serializers) throws IOException,
             JsonProcessingException {
         /*
-         * Serialization of timezone data is unwanted (not ISO). Offset is kept, timezone info is thrown away here.
-         *
-         * Keeping timezone info is a new feature which needs to be implemented.
+         * Serialization of timezone data is disabled by default, but can be turned on by enabling
+         * SerializationFeature.WRITE_DATES_WITH_ZONE_ID
          */
-        gen.writeFieldName(DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(value));
+        if (serializers.getConfig().isEnabled(SerializationFeature.WRITE_DATES_WITH_ZONE_ID)) {
+            gen.writeFieldName(DateTimeFormatter.ISO_ZONED_DATE_TIME.format(value));
+        } else {
+            gen.writeFieldName(DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(value));
+        }
     }
 
 }
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/WriteZoneIdTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/WriteZoneIdTest.java
@@ -5,9 +5,11 @@ import static org.junit.Assert.assertEquals;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.HashMap;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.fasterxml.jackson.datatype.jsr310.MockObjectConfiguration;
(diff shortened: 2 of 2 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/key/ZonedDateTimeKeySerializer.java
```diff
@@ -25,7 +26,11 @@ public void serialize(ZonedDateTime value, JsonGenerator gen, SerializerProvider
          *
          * Keeping timezone info is a new feature which needs to be implemented.
```
kupci: So now that it is implemented, the comment can be removed or changed, perhaps adding an @since?

kupci: And maybe a question for @cowtowncoder but should this go in 2.10.1 or 2.11, and then master?

cowtowncoder:
Tough question on versioning. If we wanted to follow SemVer strictly, I think it should be included in 2.11 as there is behavioral change that might not be considered a bug (or it could be, hence ambiguity).
But 2.10.0 is just out -- if we had, say, 2.10.8, I'd say 2.11.

So: @kupci what do you think? I think I am ok with either 2.10.x or 2.11.

cowtowncoder:
So: I think patch itself is fine. And the only process thing here is that unless I have already asked and received CLA:

that would be needed now (just once; no need to file for more contributions)

Thank you @vetler (Kiitos!) for the contribution; looking forward to merging it once we figure out right branch.

kupci:
Given this requires the `SerializationFeature.WRITE_DATES_WITH_ZONE_ID`, which it is now working via @vetler 's fix, **2.10.x** seems valid. In other words, this is a bug fix, so fits with 

> PATCH version when you make backwards compatible bug fixes.

cowtowncoder:
Ok, then we'll go with 2.10.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/key/ZonedDateTimeKeySerializer.java
vetler: Changed the comment, not sure where I should add the `@since`... or just remove it entirely?

vetler:
> So: I think patch itself is fine. And the only process thing here is that unless I have already asked and received CLA:
> 
> https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf

Where do I send it?

vetler:
Build failure seems unrelated to changes?

kupci:
Yes, seems unrelated to changes. I wonder if there are still Sonatype issues, as this looks like the same error as previously with the 2.10 release. There's a ticket with Sonatype opened already.

> Checksum validation failed, expected 088cd9f174c370442f12200765610c3a98bfb69e but is ae19b6d3ac494b56b7134f261d60b6d7aa9e89d2 -> [Help 1]

kupci:
And here's where to send the CLA (copying from another PR): 

> Simplest way is usually to print, fill & sign, scan, email to info at fasterxml dot com.

cowtowncoder:
Sonatype fixed the problem with `jackson-annotations`, but it is possible to have a cached copy -- so it may make sense to `rm` either whole Maven cache, or at least:

    rm ~/.m2/repository/com/fasterxml/jackson/core/jackson-annotations/

@vetler yes, send to `info` at fasterxml dot com

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/key/ZonedDateTimeKeySerializer.java
kupci: Comment looks good to me.

cowtowncoder: Ok to remove / not add `@since` in this case.

As I said, 2.10 is ok: so, either base off 2.10 and I can merge forward. Or change in `master` and I can cherry-pick back. Slightly easier first way (start with 2.10) but not a big deal.

cowtowncoder:
CLA received, proceed with merging.

PR #143, opened 2019-10-09
Fix broken links to wiki from README's

mkows:
Currently links to `wiki` found at the bottom of the following README's:
- https://github.com/FasterXML/jackson-modules-java8/tree/master/datatypes
- https://github.com/FasterXML/jackson-modules-java8/tree/master/datetime
- https://github.com/FasterXML/jackson-modules-java8/tree/master/parameter-names

are all broken as they lead to https://github.com/FasterXML/jackson-modules-java8/blob/wiki.

This PR changes target url to https://github.com/FasterXML/jackson-modules-java8/wiki
as can be seen e.g. here https://github.com/mkows/jackson-modules-java8/tree/master/datatypes (note that my repo fork does not have a wiki set up).

### DIFF
--- a/datatypes/README.md
+++ b/datatypes/README.md
@@ -73,4 +73,4 @@ System.out.println(withEmailJson);
 
 ## More
 
-See [Wiki](../../wiki) for more information (javadocs, downloads).
+See [Wiki](../../../wiki) for more information (javadocs, downloads).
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Thanks, looks good -- I haven't made up my mind on relative vs absolute, as both have their benefits/drawbacks. I tend to like relative references more but probably not big deal either way.

Issue #138, opened 2019-09-16
Prevent deserialization of "" as `null` for `Duration`, `Instant`, `LocalTime`, `OffsetTime`, and `YearMonth` in "strict" (non-lenient) mode


kupci:
For the java.time classes Duration, Instant, LocalTime, OffsetTime and YearMonth, an empty string is mapped to null by the associated deserializer, for example DurationDeserializer, InstantDeserializer, etc. So something like {"date": null} can't be distinguished from {"date": ""}. From the view of a strict API, the latter is an error.

It would be good to add some feature to treat empty string as invalid format, so that the associated deserializer (DurationDeserializer, etc) could report this as an error instead of mapping to null. This capability should be noted in the documentation.

This issue is related to issue #114 which was fixed for LocalDate and LocalDateTime. However, the fix is still required for the java.time types listed here.

cowtowncoder:
Ok, merging to `master` was some work.. .but is done.

I did however uncover one concern: #68 fix was undone by this change (2 tests broke). I think I will just comment out those tests on short term, but need to figure how these settings should work together.

kupci:
Cool, and I will take a look at #68. Lenient is set by default so empty string should pass through, so probably something else..

cowtowncoder:
I think what happened was more along the lines that I had added new handling in `master`, but due to changes from 2.11, merging, those are overridden. It's fine as nothing is released, and more a question of what should be the overall algorithm for determining acceptance. And not so much a bug introduced.

Issue #68, opened 2018-04-10
Parsing of `ZoneId` should respect `ALLOW_COERCION_OF_SCALARS` wrt empty String


cowtowncoder:
(note, moved from: https://github.com/FasterXML/jackson-datatype-jsr310/issues/101)

So: while empty String is legal to accept when `DeserializationFeature. ALLOW_COERCION_OF_SCALARS ` is enabled, it should not be accepted if it is disabled.

cowtowncoder:
Oh. Actually, no. That is not what `ACCEPT_EMPTY_STRING_AS_NULL_OBJECT` means; it only applies to structured Objects, not scalars.

cowtowncoder:
Now calls `StdDeserializer._coerceEmptyString(...)` to handle checks in standard way.
May need to improve in future from databind side, to allow per-type and/or per-property overrides, but first things is to get parity with JDK standard scalar types.

PR #145, opened 2019-10-19
restore fix for parsing of ZoneId that conflicted with changes for issue 138

kupci:
The empty string handling changes for issue #138 conflicted with the fix for issue #68, so this is to restore the previous fix, where parsing of `ZoneId` should respect `ALLOW_COERCION_OF_SCALARS` wrt empty String


cowtowncoder:
Thanks!

Issue #148, opened 2019-10-29
Allow strict `LocalDate` parsing


agusevas:
Per default `LocalDateDeserializer` deserializes local dates using `DateTimeFormatter.ISO_LOCAL_DATE` formatter. This formatter uses `ResolverStyle.STRICT`. 

```
    /**
     * Style to resolve dates and times strictly.
     * <p>
     * Using strict resolution will ensure that all parsed values are within
     * the outer range of valid values for the field. Individual fields may
     * be further processed for strictness.
     * <p>
     * For example, resolving year-month and day-of-month in the ISO calendar
     * system using strict mode will ensure that the day-of-month is valid
     * for the year-month, rejecting invalid values.
     */
    STRICT,
```

If I use `@JsonFormat` (i.e. `@JsonFormat(pattern = "yyyy-MM-dd", lenient = OptBoolean.FALSE)`) the created DateTimeFormatter uses `ResolverStyle.SMART`. It accepts invalid dates like `2019-11-31`. 
The default formatter does not accept such dates and throws an error.

To make the behaviour consistent, I suggest setting the `DateTimeFormatter` to use `ResolverStyle.STRICT` if the lenient attribute is set to `false`.

cowtowncoder:
Sounds reasonable, thank you for suggesting this!

samwill:
I'd like to work on this!

cowtowncoder:
@samwill sounds good to me.

samwill:
Added a PR with a failing test. I think I know how to fix it now; given no hiccups I should have a fix PR tonight or tomorrow morning :crossed_fingers:

cowtowncoder:
@samwill Thank you again for the fix! Will be released in 2.11.0.

PR #149, opened 2019-10-30
Test for strict localdate jsonformat #148

samwill:


samwill:
> `SMART` which of course isn't supported here - or is it?

SMART is the default actually, which I think is still used even when lenient = true currently, so LENIENT is the one that's not supported. I think there should definitely be a new issue to discuss if that should change, currently the behavior is to silently change the day-of-month to the last valid one for the month given, as long as the day-of-month value given is 31 or less. 

As for other types, now that I've gotten into it a bit, looks like this will probably impact anything extending the `JSR310DateTimeDeserializerBase` and has a concept of invalid values.

kupci:
Merged! For the ResolverStyle settings (lenient/strict/smart), what I was thinking of was how to map the Oracle behavior to the Jackson settings (lenient/strict, and various `ALLOW` flags). @cowtowncoder perhaps a discussion for the Jackson [developer list](https://groups.google.com/forum/?nomobile=true#!forum/jackson-dev)? Or maybe I'm over-analyzing things :)

cowtowncoder:
Yes, this gets into that bigger question. My initial reaction is that SMART/LENIENT distinction is quite specific to Java 8 date/time, so it'd go as a further module section to sub-divide "lenient" (Jackson-concept). Whether default ought to be one or the other is more difficult question, but from compatibility perspective I would probably go with setting that is being used now... which sounds like it is `SMART`.

With 3.0 there is possibility that date/time might yet get merged with databind, which would complicate things. But as of now, it is still separate module unlike java8 datatype (Optional types) and parameter names which have been folded in.

As to best place to discuss this: mailing list might work, but maybe the best starting point actually would be to see JSTEP-5 outline on:

https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP

and create a document proposing desired behavior. This can be aimed at 3.0, or later 2.x, or (as is case with JSTEP-3, for example), some combination, progress towards 3.0 with some parts in 2.x.

JSTEP idea is quite new and still experimental, but I like the idea of light-weight documentation, planning, to share ideas that are "bigger" than single issue, but where participation via email/chat may be too transient.

PR #164, opened 2020-02-27
Issue/#116 - Removes the fraction part based on the SerializationFeature WRITE_TIMESTAMP_WITHOUT_FRACTION

### LINKED ISSUE #116: Allow Instant to be serialized as epochSecond without the fraction part
I have an `Instant` field that represents an epoch timestamp in seconds. 

```
    @JsonProperty("registered_at")
    @JsonFormat(shape = JsonFormat.Shape.NUMBER)
    private Instant registeredAt;
```

the mapper is configured this way 

```
    .featuresToDisable(ADJUST_DATES_TO_CONTEXT_TIME_ZONE,
                       FAIL_ON_UNKNOWN_PROPERTIES,
                       WRITE_DATES_AS_TIMESTAMPS,
                       SORT_PROPERTIES_ALPHABETICALLY)
```

However when this gets serialized as

```
"registered_at" : 1420324047.000000000
```

By looking at the code there's no way to serialize this value as epoch seconds using the standard mechanism.

`InstantSerializerBase.serialize(T value, JsonGenerator generator, SerializerProvider provider)`

```
        if (useTimestamp(provider)) {
            if (useNanoseconds(provider)) {
                generator.writeNumber(DecimalUtils.toBigDecimal(
                        getEpochSeconds.applyAsLong(value), getNanoseconds.applyAsInt(value)
                ));
                return;
            }
            generator.writeNumber(getEpochMillis.applyAsLong(value));
            return;
        }
```

The only option is to use :

```
    @JsonGetter("registered_at")
    public long getRegisteredAtEpoch() {
        return registeredAt.getEpochSecond();
    }
```

It could be nice if there could be way to tell jackson to avoid serializing the fraction part.

sashikanthr:
This fixes #116 

We have added a new feature to in SerializationFeature in the databind package to avoid the fraction part. This has a separate pull request.


sashikanthr:
Since this is updating the databind module, the build will fail. We couldnt find information on how to update changes in multiple modules.

sashikanthr:
@kupci May I know if we can address your comments and update the PR? Will it be still relevant?

kupci:
@sashikanthr I think the issue is still relevant, but it might be easier to create a new PR. I think your original PR highlighted the need for a better overall solution for handling issues like these. 

One thought (refreshing myself on the issue) was to use the `Shape.*` but the difficulty there was both milliseconds and epoch seconds would be `Shape.NUMBER_INT`.

However, on that subject @cowtowncoder describes some initial thoughts / Idea-lets here:

> would it make sense to add unit (or timeUnit) property for @JsonFormat?
> 
> On one hand, it is bit datatype-specific, but there is one big benefit: ability to set format defaults, and overrides (both programmatically and via annotations) would make implementation relatively easy
> 
https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-5

One question I have is whether we can simply add more some sub-types, i.e. we already have `Shape.NUMBER`, `Shape.NUMBER_FLOAT` and `Shape.NUMBER_INT`, so could we have the new types following the Java convention, say `Shape.NUMBER_INT_MILLISECONDS`? And do we need to distinguish between epoch seconds, or would `Shape.NUMBER_INT_SECONDS` work, following the Java convention here:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html

cowtowncoder:
So, `CoercionConfig` determines reading-side only (deserialization), on whether to accept alternate, secondary input shapes: like whether `boolean` property can be read from (JSON) integral number value.
It probably will not be quite as useful here because of this, unfortunately.

kupci:
Thanks @cowtowncoder , that worked: I can get see the "Edit" button on the new-ideas wiki.

kupci:
Closing this one as per comments above, the direction for changes like this is to implement via a more long-term solution. Details here: https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-5

Issue #166, opened 2020-03-18
Cannot deserialize `OffsetDateTime.MIN` or `OffsetDateTime.MAX` with `ADJUST_DATES_TO_CONTEXT_TIME_ZONE` enabled


morth:
This probably relates to #124.

I cannot deserialize `OffsetDateTime.MIN` and `OffsetDateTime.MAX`.

With `jackson-databind` and `jackson-datatype-jsr310` (both 2.10.3) on my class path the following works:
```
ObjectMapper o = new ObjectMapper();
o.findAndRegisterModules();
o.readValue("\"" + OffsetDateTime.now().toString() + "\"", OffsetDateTime.class);
```
The following does __not__ work:
```
// (OffsetDateTime.MIN is -999999999-01-01T00:00+18:00)
o.readValue("\"" + OffsetDateTime.MIN.toString() + "\"", OffsetDateTime.class);
// Or:
// (OffsetDateTime.MAX is +999999999-12-31T23:59:59.999999999-18:00)
o.readValue("\"" + OffsetDateTime.MAX.toString() + "\"", OffsetDateTime.class);
```
Removing the offset or shifting the time by the respective amount of hours seems to work again:
```
o.readValue("\"-999999999-01-01T00:00+18:00\"", OffsetDateTime.class); // doesn't work
o.readValue("\"-999999999-01-01T18:00+18:00\"", OffsetDateTime.class); // works
o.readValue("\"-999999999-01-01T00:00+00\"", OffsetDateTime.class); // works
```

kupci:
Ok, thanks, this does seem to be an error.

cowtowncoder:
I wonder if code has assumption that values, but in UTC, would be limits? And thereby actual min/max would be considered to be outside, since normalized date (once value normalized to UTC) would then fall outside numeric values, in both cases.

morth:
I digged a little bit deeper into the issue and the problem seems to be the `ADJUST_DATES_TO_CONTEXT_TIME_ZONE` deserialization feature.

[...]

PR #172, opened 2020-05-15
#166: Only normalize OffsetDateTimes when they are normalizable

morth:
Possible fix for #166. This commit adds a check before the OffsetDateTime adjustment of ADJUST_DATES_TO_CONTEXT_TIME_ZONE that ensures that the OffsetDateTime is normalizeable given a specific time zone offset. When it is not, normalization is skipped and the un-adjusted OffsetDateTime is returned.


cowtowncoder:
+1 for suggestion on static member. One other suggestion: if easy, rebasing against 2.11 (if considered safe change) or 2.12 (if there is risk) would be nice. That can then be merged forward 2.11 -> 2.12 -> master (3.0). Not a big deal as cherry-picking back from `master` is probably easy as well.

morth:
@kupci @cowtowncoder I implemented your suggestions and fixed this in 2.11 in #173.

@kupci 'Ex' was for "excluded", since it was the first value outside of the boundary.

PR #173, opened 2020-05-16
Only normalize OffsetDateTimes when they are normalizable

morth:
Fix for #166. This commit introduces static constants for the minimum and the maximum OffsetDateTimes that are adjustable to all possible time zones in InstantDeserializer and adds a check that an OffsetDateTime lies in these boundaries before it is adjusted.


cowtowncoder:
LGTM; @kupci what do you think?

@morth Thank you for this PR & apologies for delays. Hoping to get this merged soon -- just one more thing I need (... unless I asked for and it got it already), CLA. Found here:

and only needs to be filed once. Easiest way to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I have that (and yay from mr kupci), happy to merge this to be included in 2.11.1!

kupci:
Moved these fixes, with a few changes, over to another PR so closing this one.

PR #187, opened 2020-10-12
Support for jsonformat in duration deserializer

### LINKED ISSUE #184: `DurationDeserializer` should use `@JsonFormat.pattern` (and config override) to support configurable `ChronoUnit`
When deserializing `DurationDeserializer` treats all durations of number values as seconds and delegates string values to `Duration.parse`. 

## Versions

```
com.fasterxml.jackson.core:jackson-core:jar:2.11.2
com.fasterxml.jackson.core:jackson-databind:jar:2.11.2
com.fasterxml.jackson.datatype:jackson-datatype-jdk8:jar:2.11.2
com.fasterxml.jackson.datatype:jackson-datatype-jsr310:jar:2.11.2
com.fasterxml.jackson.core:jackson-annotations:jar:2.11.2
```

## Example

```java
class A {
    @JsonFormat(pattern = "h")
    Duration dur;

    public static void main(String[] args) throws Exception {
        JsonMapper m = JsonMapper.builder()
                .build();
        A a0 = m.readValue("{\"dur\":2}".getBytes(), A.class);
        System.out.println(a0.dur.equals(Duration.ofHours(2))); // should print "true" but prints "false

        A a1 = m.readValue("{\"dur\":\"2\"}".getBytes(), A.class);
        System.out.println(a1.dur.equals(Duration.ofHours(2))); // should print "true" but throws Exception
    }
}
```

The Exception thrown by the second `readValue` call:

```
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.Duration` from String "2": Failed to deserialize java.time.Duration: (java.time.format.DateTimeParseException) Text cannot be parsed to a Duration
 at [Source: (byte[])"{"dur":"2"}"; line: 1, column: 8] (through reference chain: *****)

	at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:67)
	at com.fasterxml.jackson.databind.DeserializationContext.weirdStringException(DeserializationContext.java:1702)
	at com.fasterxml.jackson.databind.DeserializationContext.handleWeirdStringValue(DeserializationContext.java:947)
	at com.fasterxml.jackson.datatype.jsr310.deser.JSR310DeserializerBase._handleDateTimeException(JSR310DeserializerBase.java:129)
	at com.fasterxml.jackson.datatype.jsr310.deser.DurationDeserializer.deserialize(DurationDeserializer.java:92)
	at com.fasterxml.jackson.datatype.jsr310.deser.DurationDeserializer.deserialize(DurationDeserializer.java:43)
	at com.fasterxml.jackson.databind.deser.impl.FieldProperty.deserializeAndSet(FieldProperty.java:138)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:293)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:156)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4524)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3527)
        ...
Caused by: java.time.format.DateTimeParseException: Text cannot be parsed to a Duration
	at java.base/java.time.Duration.parse(Duration.java:417)
	at com.fasterxml.jackson.datatype.jsr310.deser.DurationDeserializer.deserialize(DurationDeserializer.java:90)
	... 70 more
```

Even though `JsonFormat` doesn't explicitly mention any `jsr310` classes in its documentation they should still work with it. Especially since `DurationDeserializer` already respects `JsonFormat#lenient`.

obarcelonap:
Implementation based on Duration::of(long,TemporalUnit). Closes #184

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
@@ -34,6 +34,13 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import java.time.DateTimeException;
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
 
 
 /**
@@ -49,8 +56,17 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
 
     public static final DurationDeserializer INSTANCE = new DurationDeserializer();
 
-    private DurationDeserializer()
-    {
+    /**
+     * Since 2.12
+     * When set, integer values will be deserialized using the specified unit. Using this parser will tipically
+     * override the value specified in {@link DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS} as it is
+     * considered that the unit set in {@link JsonFormat#pattern()} has precedence since is more specific.
+     *
+     * @see [jackson-modules-java8#184] for more info
+     */
+    private DurationUnitParser _durationUnitParser;
+
+    private DurationDeserializer() {
         super(Duration.class);
     }
 
@@ -59,6 +75,11 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
      */
     protected DurationDeserializer(DurationDeserializer base, Boolean leniency) {
         super(base, leniency);
+    }
+
+    protected DurationDeserializer(DurationDeserializer base, DurationUnitParser durationUnitParser) {
+        super(base, base._isLenient);
+        _durationUnitParser = durationUnitParser;
     }
 
     @Override
@@ -79,8 +100,17 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
                     deser = deser.withLeniency(leniency);
                 }
             }
+            if (format.hasPattern()) {
+                deser = DurationUnitParser.from(format.getPattern())
+                        .map(deser::withPattern)
+                        .orElse(deser);
+            }
         }
         return deser;
+    }
+
+    private DurationDeserializer withPattern(DurationUnitParser pattern) {
+        return new DurationDeserializer(this, pattern);
     }
 
     @Override
@@ -92,7 +122,11 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
                 BigDecimal value = parser.getDecimalValue();
                 return DecimalUtils.extractSecondsAndNanos(value, Duration::ofSeconds);
             case JsonTokenId.ID_NUMBER_INT:
-                return _fromTimestamp(context, parser.getLongValue());
+                long intValue = parser.getLongValue();
+                if (_durationUnitParser != null) {
+                    return _durationUnitParser.parse(intValue);
+                }
+                return _fromTimestamp(context, intValue);
             case JsonTokenId.ID_STRING:
                 return _fromString(parser, context, parser.getText());
             // 30-Sep-2020, tatu: New! "Scalar from Object" (mostly for XML)
@@ -103,9 +137,9 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
                 // 20-Apr-2016, tatu: Related to [databind#1208], can try supporting embedded
                 //    values quite easily
                 return (Duration) parser.getEmbeddedObject();
-                
+
             case JsonTokenId.ID_START_ARRAY:
-            	return _deserializeFromArray(parser, context);
+                return _deserializeFromArray(parser, context);
         }
         return _handleUnexpectedToken(context, parser, JsonToken.VALUE_STRING,
                 JsonToken.VALUE_NUMBER_INT, JsonToken.VALUE_NUMBER_FLOAT);
@@ -140,5 +174,34 @@ public class DurationDeserializer extends JSR310DeserializerBase<Duration>
             return Duration.ofSeconds(ts);
         }
         return Duration.ofMillis(ts);
+    }
+
+    protected static class DurationUnitParser {
+        final static Set<ChronoUnit> PARSEABLE_UNITS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+                ChronoUnit.NANOS,
+                ChronoUnit.MICROS,
+                ChronoUnit.MILLIS,
+                ChronoUnit.SECONDS,
+                ChronoUnit.MINUTES,
+                ChronoUnit.HOURS,
+                ChronoUnit.HALF_DAYS,
+                ChronoUnit.DAYS
+        )));
+        final TemporalUnit unit;
+
+        DurationUnitParser(TemporalUnit unit) {
+            this.unit = unit;
+        }
+
+        Duration parse(long value) {
+            return Duration.of(value, unit);
+        }
+
+        static Optional<DurationUnitParser> from(String unit) {
+            return PARSEABLE_UNITS.stream()
+                    .filter(u -> u.name().equals(unit))
+                    .map(DurationUnitParser::new)
+                    .findFirst();
+        }
     }
 }
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationUnitParserTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationUnitParserTest.java
@@ -1,0 +1,53 @@
+package com.fasterxml.jackson.datatype.jsr310.deser;
+
+import com.fasterxml.jackson.datatype.jsr310.deser.DurationDeserializer.DurationUnitParser;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
+import java.util.Collection;
+import java.util.Optional;
+
+import static java.util.Arrays.asList;
+import static java.util.Optional.of;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class DurationUnitParserTest {
+
+    private final String stringPattern;
+    private final TemporalUnit temporalUnit;
+
+    public DurationUnitParserTest(String stringPattern, TemporalUnit temporalUnit) {
+        this.stringPattern = stringPattern;
+        this.temporalUnit = temporalUnit;
+    }
+
+    @Test
+    public void shouldMapToTemporalUnit() {
+        Optional<DurationUnitParser> durationPattern = DurationUnitParser.from(stringPattern);
+
+        assertEquals(of(temporalUnit), durationPattern.map(dp -> dp.unit));
+    }
+
+    @Parameters
+    public static Collection<Object[]> testCases() {
+        return asList(
+                asArray("NANOS", ChronoUnit.NANOS),
+                asArray("MICROS", ChronoUnit.MICROS),
+                asArray("MILLIS", ChronoUnit.MILLIS),
+                asArray("SECONDS", ChronoUnit.SECONDS),
+                asArray("MINUTES", ChronoUnit.MINUTES),
+                asArray("HOURS", ChronoUnit.HOURS),
+                asArray("HALF_DAYS", ChronoUnit.HALF_DAYS),
+                asArray("DAYS", ChronoUnit.DAYS)
+        );
+    }
+
+    private static Object[] asArray(Object... values) {
+        return values;
+    }
+}
--- a/datetime/README.md
+++ b/datetime/README.md
@@ -47,6 +47,20 @@ times but are supported with this module nonetheless.
 [`LocalDateTime`](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html), and
 [`OffsetTime`](https://docs.oracle.com/javase/8/docs/api/java/time/OffsetTime.html), which cannot portably be converted to
 timestamps and are instead represented as arrays when `WRITE_DATES_AS_TIMESTAMPS` is enabled.
+* [`Duration`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html), which unit can be configured in `JsonFormat` using a subset of [`ChronoUnit`](https://docs.oracle.com/javase/8/docs/api/java/time/temporal/ChronoUnit.html) as `pattern`. 
+As the underlying implementation is based on `Duration::of` supported units are: `NANOS`, `MICROS`, `MILLIS`, `SECONDS`, `MINUTES`, `HOURS`, `HALF_DAYS` and `DAYS`.
+For instance:
+
+    ```java
+    @JsonFormat(pattern="MILLIS")
+    long millis;
+    
+    @JsonFormat(pattern="SECONDS")
+    long seconds;
+    
+    @JsonFormat(pattern="DAYS")
+    long days;
+    ```
 
 ## Usage
 
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserTest.java
@@ -1,20 +1,7 @@
 package com.fasterxml.jackson.datatype.jsr310.deser;
 
-import java.math.BigInteger;
-import java.time.Duration;
-import java.time.temporal.TemporalAmount;
-import java.util.Map;
-
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.core.type.TypeReference;
-import org.junit.Test;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -22,12 +9,33 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.exc.MismatchedInputException;
 import com.fasterxml.jackson.datatype.jsr310.MockObjectConfiguration;
 import com.fasterxml.jackson.datatype.jsr310.ModuleTestBase;
+import org.junit.Test;
+
+import java.math.BigInteger;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalAmount;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 public class DurationDeserTest extends ModuleTestBase
 {
     private final ObjectReader READER = newMapper().readerFor(Duration.class);
 
     private final TypeReference<Map<String, Duration>> MAP_TYPE_REF = new TypeReference<Map<String, Duration>>() { };
+
+    final static class Wrapper {
+        public Duration value;
+
+        public Wrapper() { }
+        public Wrapper(Duration v) { value = v; }
+    }
+
 
     @Test
     public void testDeserializationAsFloat01() throws Exception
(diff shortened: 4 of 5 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -79,20 +88,38 @@ protected DurationDeserializer withLeniency(Boolean leniency) {
...
+                    return pattern.parse(decValue.longValue());
+                }
+                return DecimalUtils.extractSecondsAndNanos(decValue, Duration::ofSeconds);
             case JsonTokenId.ID_NUMBER_INT:
-                return _fromTimestamp(context, parser.getLongValue());
+                long intValue = parser.getLongValue();
+                if (pattern != null) {
+                    return pattern.parse(intValue);
+                }
+                return _fromTimestamp(context, intValue);
```
kupci, line 122: So the new functionality will override the flag `DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS` if already set... guessing that is ok, though to be safer you might want to adhere to the flag if both are set, and there have a conflict, i.e they don't specify nanos in the pattern. The documentation should be updated too.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -121,13 +148,18 @@ protected Duration _fromString(JsonParser parser, DeserializationContext ctxt,
...
+
         // 30-Sep-2020: Should allow use of "Timestamp as String" for
         //     some textual formats
         if (ctxt.isEnabled(StreamReadCapability.UNTYPED_SCALARS)
                 && _isValidTimestampString(value)) {
             return _fromTimestamp(ctxt, NumberInput.parseLong(value));
         }
 
+        if (pattern != null) {
+            return pattern.parse(NumberInput.parseLong(value));
```
kupci, line 160: Exception handling tested?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -49,6 +53,11 @@
 
     public static final DurationDeserializer INSTANCE = new DurationDeserializer();
 
+    /**
+     * Since 2.12
+     */
+    private DurationPattern pattern;
```
kupci, line 59: Recommend adding some JavaDoc here to describe the new variable. Also suggest naming it `_durationPattern` for consistency with other private vars and to be a little more specific.

kupci (commented):
Took a brief look, looks good, may have some thoughts later. One thing that will be nice about this change is that it looks like .._eventually_ one of the DeserializationFeature flags can be deprecated.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -79,20 +88,38 @@ protected DurationDeserializer withLeniency(Boolean leniency) {
                     deser = deser.withLeniency(leniency);
                 }
             }
+            if (format.hasPattern()) {
+                this.pattern = DurationPattern.from(format.getPattern())
+                        .orElse(null);
```
kupci, line 93: Wondering if this should follow the "pattern" of other configuration settings, and set pattern in the super class, and then call a method like the following:
`deser = deser.withPattern(format.getPattern());`

This would set the groundwork for other types adding support for 'pattern's, which might help with some tricky bugs. Or is this duplicating work in date-time-config? Thoughts @cowtowncoder ?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: Done

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: I generalized the handling of `DateTimeException` in the base class and tested the code.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: That could make sense, but I think is a bigger topic so I Ieft the disussion up to you :)

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: I think not honouring the flag is ok since the `@JsonFormat` annotation has been used explicitly may have more precedence. Actually that flag is not well used in this serializer because it uses `ofSeconds` while the flag is about nanos.

obarcelonap: BTW, what docs should I update?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: But in any case, adding a unit in JsonFormat is not a conflict with the current change and it can be added later on.

obarcelonap:
Did the changes in form of `--fixup` I will rebase once is ok from your side

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
kupci: I think a line here, to add Duration to the list of exceptions, and mention that it supports "pattern", would be helpful (in the section after the line below)

> Some exceptions to this standard serialization/deserialization rule:

https://github.com/FasterXML/jackson-modules-java8/tree/master/datetime

kupci (commented):
Thanks for the changes, they look good to me. Approved - assuming you'll get the documentation in. If they look good to @cowtowncoder too, you'll need to send him the CLA if you haven't already.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -49,6 +53,11 @@
 
     public static final DurationDeserializer INSTANCE = new DurationDeserializer();
 
+    /**
+     * Since 2.12
```
kupci, line 57: Suggest adding complete JavaDoc, see this for a template you could use: https://github.com/FasterXML/jackson-modules-java8/blob/eeb2b67de6fdc975d9bde88d32b2c1c83a18d710/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java#L22

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: Done

obarcelonap:
> Thanks for the changes, they look good to me. Approved - assuming you'll get the documentation in. If they look good to @cowtowncoder too, you'll need to send him the CLA if you haven't already.

I haven't sent the CLA, where can I find it?

Just rebased on top of 2.12 branch, added missing documentation and squashed all the commits.

cowtowncoder:
CLA can be found from `jackson` repo:

obarcelonap:
Just sent the CLA by email.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -129,9 +164,13 @@ protected Duration _fromString(JsonParser parser, DeserializationContext ctxt,
...
         try {
+            if (_durationUnitParser != null) {
+                return _durationUnitParser.parse(NumberInput.parseLong(value));
+            }
+
             return Duration.parse(value);
-        } catch (DateTimeException e) {
-            return _handleDateTimeException(ctxt, e, value);
+        } catch (NumberFormatException | DateTimeException e) {
+            return _handleWeirdStringValue(ctxt, e, value);
```
kupci, line 173: @cowtowncoder  Ok to use same exception handler method, for a number? Or should be _handleWeirdNumberValue(...)?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -79,20 +97,36 @@ protected DurationDeserializer withLeniency(Boolean leniency) {
                     deser = deser.withLeniency(leniency);
                 }
             }
+            if (format.hasPattern()) {
+                deser = DurationUnitParser.from(format.getPattern())
```
kupci, line 101: @obarcelonap Are you sure this preserves the `leniency` setting from above? Or should follow same convention as above, i.e.
`deser = deser.withPattern(..) `

kupci (commented):
Added a couple more comments/2nd (or third) thoughts, since figuring easier to fix here than later..

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: I think is ok to use same handler because at the end the original value that is on processing is a `String`, so it matches the handler purpose.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: Yes it is, `withPattern` is referenced when there is a value using the `Optional` monad.
As you already pointed, the non functional approach, would be:
```
deser = deser.withPattern(DurationUnitParser.from(format.getPattern()).orElse(null))
```
Let me know if you prefer this second approach.

obarcelonap (commented):
> Added a couple more comments/2nd (or third) thoughts, since figuring easier to fix here than later..

Just replied, I think is mainly code style but I can change the code if it fits better in this repo codebase. Let me know

cowtowncoder:
Oh. Hmm. I should have looked at this earlier. I'll have to think how I feel about specifying new patterns -- but it is good that at least it is based on `ChronoUnit`; and the limitation of Java annotations prevents doing much more than matching it explicitly.
I'll go over code and hopefully will be able to merge this soon.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -79,20 +97,36 @@ protected DurationDeserializer withLeniency(Boolean leniency) {
...
     public Duration deserialize(JsonParser parser, DeserializationContext context) throws IOException
     {
         switch (parser.currentTokenId())
         {
             case JsonTokenId.ID_NUMBER_FLOAT:
-                BigDecimal value = parser.getDecimalValue();
-                return DecimalUtils.extractSecondsAndNanos(value, Duration::ofSeconds);
+                BigDecimal decValue = parser.getDecimalValue();
+                if (_durationUnitParser != null) {
+                    return _durationUnitParser.parse(decValue.longValue());
```
cowtowncoder, line 121: This would be lossy... I wonder if the right handling would be throw an exception; or perhaps require a coercion check? (2.12 has `CoercionConfig` that could be checked to see if "floating-point -> integer" coercion is legal; by default it is, but can prevent.
This assuming that it is not possible to use fractions in meaningul way... is there?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DeserializerBase.java
```diff
@@ -141,6 +141,13 @@ protected boolean _isValidTimestampString(String str) {
     @SuppressWarnings("unchecked")
     protected <R> R _handleDateTimeException(DeserializationContext context,
               DateTimeException e0, String value) throws JsonMappingException
+    {
+        return _handleWeirdStringValue(context, e0, value);
+    }
+
+    @SuppressWarnings("unchecked")
+    protected <R, E extends RuntimeException> R _handleWeirdStringValue(DeserializationContext context,
```
cowtowncoder, line 149: No need to add type parameter `E`, I think, ok to just pass `Exception`, I think.

cowtowncoder:
Ok I think code looks pretty good on its own; I can make small changes myself if need be.

So the first question I have is whether decimal/floating-point numbers (vs integers) should be handled -- and it looks like only "ofSeconds()" handles two parts, although in theory we could calculate "1.5 hours" into 90 minutes just fine... but it gets complicated and I wonder would it not be better to simply throw an exception to let user know that fractions are not allowed for anything but seconds? 

Second part is wrt serialization: should format not also be considered when serializing?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -141,4 +180,23 @@ protected Duration _fromTimestamp(DeserializationContext ctxt, long ts) {
...
+            this.unit = unit;
+        }
+
+        Duration parse(long value) {
+            return Duration.of(value, unit);
+        }
+
+        static Optional<DurationUnitParser> from(String unit) {
+            return Stream.of(ChronoUnit.values())
+                    .filter(u -> u.toString().equalsIgnoreCase(unit))
```
cowtowncoder, line 197: I think we should not actually allow case-insensitive match, but require upper-case, to exactly match `ChronoUnit`.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
```diff
@@ -129,9 +164,13 @@ protected Duration _fromString(JsonParser parser, DeserializationContext ctxt,
         }
 
         try {
+            if (_durationUnitParser != null) {
+                return _durationUnitParser.parse(NumberInput.parseLong(value));
```
cowtowncoder, line 168: Ok, here's another potential question: should JSON Strings be allowed with pattern?
In that case we'd probably want to check `CoercionConfig` as input absolutely must be (integer) number, so finding JSON String requires coercion. Same is not true for default `Duration` which conversely must get String.

I can probably play with this a bit so above is just my thinking out aloud.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
kupci: > would it not be better to simply throw an exception to let user know that fractions are not allowed for anything but seconds?

Didn't even think of that, was assuming no fractions. So yes, my vote would be "simpler is better".

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
kupci: Limit the DurationParser to only `ID_NUMBER_INT` logic, to match the longs that are available in the `toXXX` methods of `Duration`, perhaps? (Now that I'm thinking about how this would be serialized.)
https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toDays--

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationUnitParserTest.java
```diff
@@ -0,0 +1,77 @@
...
+                asArray("HalfDays", ChronoUnit.HALF_DAYS),
+                asArray("Days", ChronoUnit.DAYS),
+                asArray("Weeks", ChronoUnit.WEEKS),
+                asArray("Months", ChronoUnit.MONTHS),
+                asArray("Years", ChronoUnit.YEARS),
+                asArray("Decades", ChronoUnit.DECADES),
+                asArray("Centuries", ChronoUnit.CENTURIES),
+                asArray("Millennia", ChronoUnit.MILLENNIA),
+                asArray("Eras", ChronoUnit.ERAS),
+                asArray("Forever", ChronoUnit.FOREVER)
```
kupci, line 57: TIL: There is a `ChronoUnit` for "forever". Wondering if we can trim down the actual realistically supported `ChronoUnit`s?

kupci:
> Second part is wrt serialization: should format not also be considered when serializing?

Hmm, glossed over this, but now I think so. The `DurationSerializer` supports milliseconds, seconds/nanoseconds, or ["string representation"](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toString--):

> A string representation of this duration using ISO-8601 seconds based representation, such as PT8H6M12.345S. 

Now in reviewing, it looks like we can trim supported units down the actual permitted `ChronoUnit`s supported by `Duration` methods, i.e. `toDays()`, `toHours()`, `toMillis()`, `toMinutes()`, `toNanos()`? 

Probably also need to test (with the serializer) edge cases like `ChronUnit.FOREVER` ;)

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
cowtowncoder: Yeah; it is easy enough to add support for `String` coercion if need be (if requested).
So let's go with that.

I might still go ahead and check for `CoercionConfig` for existing coercions -- I know "lenient" is used in some places, but I think my thinking has evolved so that there is difference between:

1. Coercion from different input value shape (like JSON String) into something that would usually be required (like integral number). For this we have CoercionConfig and a few legacy features. Leniency does not come to play as it is too coarse setting (on/off)
2. Leniency would affect handling within expected value domain -- so, for String representation of Date, whether we allow "2021-02-30", or similar (case-sensitivity option exists, too, which is yet separate)

cowtowncoder:
I think I have a few different things/ideas to discuss: will add to issue itself. I hope we can get this merged soon, but I really want to Do It Right, including backwards compatibility.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
kupci: So leave this case block ( case JsonTokenId.ID_NUMBER_FLOAT:) as-is, with no changes, and the only case block that should be changed with this PR is the one for `JsonTokenId.ID_NUMBER_INT'? Keep things simpler for now?

cowtowncoder: Well, kind of; change should be isolated in `_fromTimestamp()`, which can also be called from `_fromString()` -- for special case of XML, CSV etc (formats for which `ctxt.isEnabled(StreamReadCapability.UNTYPED_SCALARS)` returns true).
This because those formats can not reliably indicate number type (XML not without Schema, CSV not without something external), so auto-detection of integral numbers makes sense without coercion.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: since `String` is not handled anymore I reverted these changes.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: Only integers are handled now

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: Now matching is case sensitive.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationDeserializer.java
obarcelonap: Only integers respects pattern for the moment.

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/DurationUnitParserTest.java
obarcelonap: Not supported, fixed. Actually everthing greater than `DAYS` is not supported because is considered not precise. I fixed the code and added some tests.

obarcelonap:
Solved your comments guys, thanks for your feedback!

Issue #188, opened 2020-10-15
[2.12.0-rc1] Can no longer serialize using different timezones on independent OffsetDateTime fields


WolfEkk:
Hi it would seem like #185 (PR for #175) removes the possibility of producing different offset formats with the same ObjectMapper.

In the example below you will see 3 cases:
- One using an ObjectMapper without setting it's time zone.
- One using an ObjectMapper with it's time zone set to the value appropriate for the the `offset` field.
- One using an ObjectMapper with it's time zone set to the value appropriate for the the `offset2` field.

None of these options seem to work. I think it might just be necessary to honor the given object's timezone when the timezone of the ObjectMapper is not set, I would love to hear your opinion on this or if there's an alternate way to make this work I would also love to hear about it.

I've added a fourth test case which shows that deserialization is possible when using  `DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE`. This makes me think that at a higher level the change introduces asymmetry in the behavior, and I think being able to go back and forth preserving the data is important.

Please let me know if I can improve this report or assist in any way.

```
import static org.junit.Assert.assertEquals;

import com.fasterxml.jackson.annotation.*;
import com.fasterxml.jackson.annotation.JsonCreator.Mode;
import com.fasterxml.jackson.databind.*;
import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
import java.time.*;
import java.util.TimeZone;
import org.junit.Test;

public class ObjectMapperTest {

    private static final LocalDateTime LOCAL = LocalDateTime.of(2018, 11, 26, 17, 53, 22);
    private static final OffsetDateTime OFFSET = OffsetDateTime.of(LOCAL.plusSeconds(1), ZoneOffset.of("-05:00"));
    private static final OffsetDateTime OFFSET_2 = OffsetDateTime.of(LOCAL.plusSeconds(3), ZoneOffset.of("Z"));
    private static final TestPojo TEST_POJO = new TestPojo(OFFSET, OFFSET_2);
    private static final String EXPECTED_JSON = "{"
            + "\"offset\":\"2018-11-26T17:53:23-05:00\","
            + "\"offset2\":\"11/26/2018 17:53:25Z\""
            + "}";

    @Test
    public void testSerialization() throws Exception {
        ObjectMapper subject = newObjectMapper();
        String json = subject.writeValueAsString(TEST_POJO);
[...]

Issue #175, opened 2020-05-19
ObjectMapper#setTimeZone ignored by jsr-310/datetime types during serialization when using `@JsonFormat` annotation 


ErwanLeroux:
Hello, i have an issue regarding the serialization of OffsetDateTime, the timezone i set in the object mapper is not used

With the following code
```
public class POJO {
    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX")
    private OffsetDateTime offsetDateTimeWithoutTimeZone;
}
```
```
public static void setUp() {
    mapperWithTimezone = new ObjectMapper()
            .registerModule(new JavaTimeModule())
            .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
            .setTimeZone(TimeZone.getTimeZone("Europe/Paris"))
            .enable(SerializationFeature.INDENT_OUTPUT);
}
```
```
final OffsetDateTime firstJune2020AtNoonOffsetUTC = OffsetDateTime.of(
        LocalDate.of(2020, Month.JUNE, 1),
        LocalTime.NOON,
        ZoneOffset.UTC);

final POJO pojoOffsetUTC = new POJO();
pojoOffsetUTC.setOffsetDateTimeWithoutTimeZone(firstJune2020AtNoonOffsetUTC);
final String strPojoOffSetUTC = mapperWithTimezone.writeValueAsString(pojoOffsetUTC);
```
I'm expecting that the output is

```{"offsetDateTimeWithoutTimeZone" : "2020-06-01T14:00:00.000+02:00"}```

 but i get

```{"offsetDateTimeWithoutTimeZone" : "2020-06-01T12:00:00.000Z"}```

The javadoc of com.fasterxml.jackson.annotation.JsonFormat#timezone tells me that the default is the one specified in the objectMapper, but UTC is used instead.

I search a bit in the code, and i didn't see anything regarding the use of ObjectMapper#setTimezone in the method com.fasterxml.jackson.datatype.jsr310.ser.JSR310FormattedSerializerBase#createContextual

When i specify the timezone in the JsonFormat annotation, it work as intended, but i would prefer to have a global setting rather specifying it for each field.

I created a repo with a working code to illustrate my question : https://github.com/ErwanLeroux/jackson-issue

cowtowncoder:
Sounds like a bug, thank you for reporting this.
[...]

PR #190, opened 2020-10-20
Implement #188: only override timezone/offset if explicitly defined default offset/tz

cowtowncoder:
As per title, try to reduce scope of #175 using new configuration setting.


cowtowncoder:
Uhh. Wrong branch.

PR #191, opened 2020-10-20
Implement #188: only override timezone/offset if explicitly defined timezone/offset

cowtowncoder:
As per title, try to reduce scope of #175 using new configuration setting.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/InstantSerializerBase.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/InstantSerializerBase.java
@@ -16,17 +16,18 @@
 
 package com.fasterxml.jackson.datatype.jsr310.ser;
 
-import com.fasterxml.jackson.annotation.JsonFormat;
 import java.io.IOException;
-import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.Temporal;
 import java.util.function.ToIntFunction;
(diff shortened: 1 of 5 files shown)


cowtowncoder:
Ok, @kupci, @WolfEkk, does this make sense? Basically default timezone/offset will only be used to override one in date/time value if explicitly set when `ObjectMapper` built, but not if left as default (or, if default specified as `null`, i.e. remove override).
As per comment there may be need for further configurability (just need to figure out how); but for now seems like a small step we can take.

WolfEkk:
I think this is a good compromise, it leaves the opportunity use cases like mine and for the most part it preserves the behavior expected from #185.

I only wish I could have contributed the fix myself but I couldn't carve enough time out. This is the first time I reach out to you guys and I'd really like to thank you for the attention and the prompt replies and follow-ups.

I'll stick around, maybe I get the chance to contribute, even if just a little in the near future. Have a great week! Looking forward for the rc2 release :)

cowtowncoder:
@WolfEkk Sounds good, and your help here is appreciated. Date/time handling is one of trickier areas both due to somewhat complex problem domain, and due to wide set of contributions for this module with maybe sometimes different ideas. Would definitely be great if you have time to dig into code at some point here (or other modules).

cowtowncoder:
Assuming this is generally accepted, merging.

atharvabirtharia:
hi please see my code too

cowtowncoder:
@atharvabirtharia Hmmh?

Issue #94, opened 2018-12-10
Deserialization of timestamps with UTC timezone to LocalDateTime doesn't yield correct time


alankila:
Hello,

I'm using LocalDateTime to represent my timestamps on server side, but would like to transmit data in instant-like format in and out of my API, using the Z timezone. One of the advantages of doing this is that all clients treat the string as referring to the same instance, e.g. "new Date(string)" will result in same interpretation in all browsers, and built-in function such as date.toISOString() exists for formatting the string into an instant in time.

I ran into this behavior in Jackson during deserializing:

https://github.com/FasterXML/jackson-modules-java8/blob/66dce029e7c2966738251155f9c4f2c69e26a355/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserializer.java#L75

The behavior of this line is equivalent to just removing the Z character at end of timestamp and treating the rest as LocalDateTime, e.g. it will return the LocalDateTime in the UTC timezone. Wouldn't it make a lot more sense to convert UTC timestamp to LocalDateTime using system's own timezone, i.e.

    LocalDateTime.ofInstant(Instant.parse(string), ZoneId.systemDefault())

so that the time as written by JavaScript would be understood as the same time after Jackson is processing it? I don't think a lot of people expect that a LocalDateTime should be interpreted to be in UTC timezone. For most of us, it's expected to be exactly what it says on the class name: a date and time in the local timezone.

kupci:
This makes sense, and as a further confirmation that this is should be the expected behavior, if you call `now()` on a [LocalDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html) instance, it will return the 
> the current date-time from the system clock in the **default time-zone**.
> This will query the system clock in the default time-zone to obtain the current date-time. 

Though how to implement change, since would be a breaking change? There is the `ADJUST_DATES_TO_CONTEXT_TIME_ZONE` feature already (see issue #118) which uses this logic to get the Zone (if the feature is enabled):

>     private ZoneId getZone(DeserializationContext context)
>     {
>         // Instants are always in UTC, so don't waste compute cycles
>         return (_valueClass == Instant.class) ? null : context.getTimeZone().toZoneId();
>     }

[...]

PR #192, opened 2020-10-21
Fail deserialization into LocalDateTime of datetime with timezone

angelyan:
Closes https://github.com/FasterXML/jackson-modules-java8/issues/94


cowtowncoder:
Thank you @angelyan!

One thing (aside from my verifying if anyone has concerns) that I'd need before merging is CLA from:

and the usual way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I get that I should be able to merge, backport in for 2.12 (should cherry-pick just fine).
The only question I have is as to whether some handling would be needed for special case (to possibly strip out trailing `Z`), but it is quite difficult to really know what all users might be doing (kinds of unexpected or misguided usage).
So hopefully by getting this in, releasing `2.12.0-rc2`, we could get more validation.

kupci (approved):

cowtowncoder:
Merged, thanks!

Also ended up changing a bit as per comments on #94.

Issue #189, opened 2020-10-18
Support use of "pattern" (`ChronoUnit`) for `DurationSerializer` too


cowtowncoder:
(see #184 for background)

So, we should support use of `@JsonFormat(pattern = "HOURS")` (and "config override" equivalent) for writing too. As discussed on #184, there is gnarly challenge with backwards-compatibility -- change should only affect explicit pattern-using case.

Another challenge is the possibility of fractional values: given `Duration` may not be exactly divisible by desired unit: for example, we might have "90 seconds" as value, but configured to use "minutes", resulting in `1.5` (ideally).
If it is possible to calculate this accurately (enough), I'd like to do that, but I am not sure it is.

If not, I suggest that only one specific case of fractions would be supported: that of "SECONDS" as _explicit_ unit. It would be output as fractions as needed, based on underlying seconds/nanoseconds values.

cowtowncoder:
Also note that above logic should only be invoked when configured to output "as timestamp" (as number) -- "as string" uses `Duration.toString()`, which in turn produces `Period` style textual output. We should leave that aspect unchanged as it is at least quite well defined.

obarcelonap:
Hey guys, can I take this one too?
Thanks in advance

cowtowncoder:
@obarcelonap sure! I made some changes to the first one so make sure sync (some renaming). Was thinking that helper class probably should go under new `.util` package, but did not yet move there.

obarcelonap:
Hey guys just worked a bit today on this
* I needed to abstract & override couple of methods from base serializer
* The pattern only applies when `SerializationFeature.WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS` is enabled and format pattern is detected.
* Not supporting fractional values yet, I need to iterate on this. My idea is to enable fractions only when `SerializationFeature.WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS` is enabled and a format pattern is detected

Branch here in case you want to take a look https://github.com/FasterXML/jackson-modules-java8/compare/master...obarcelonap:support-pattern-for-duration-serializer?expand=1

kupci:
One comment on code below:
 ```
   public DurationSerializer(DurationSerializer base, DurationUnitConverter converter) {
        super(base, base._useTimestamp, base._useNanoseconds, base._formatter, base._shape);
        _durationUnitConverter = converter;
    }
```
[...]

PR #194, opened 2020-10-27
Support pattern for duration serializer

obarcelonap:
Does not support duration serialization with fractions (see comments in #189). I would propose to merge first this so we can iterate and add fractions feature in the future.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/DurationSerializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/DurationSerializer.java
@@ -21,14 +21,17 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 
+import com.fasterxml.jackson.databind.BeanProperty;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonSerializer;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonFormatVisitorWrapper;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonIntegerFormatVisitor;
 import com.fasterxml.jackson.databind.jsonFormatVisitors.JsonValueFormat;
 import com.fasterxml.jackson.datatype.jsr310.DecimalUtils;
+import com.fasterxml.jackson.datatype.jsr310.util.DurationUnitConverter;
 
 import java.io.IOException;
 import java.math.BigDecimal;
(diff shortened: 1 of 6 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/util/DurationUnitConverter.java
```diff
@@ -0,0 +1,76 @@
...
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static com.fasterxml.jackson.datatype.jsr310.util.DurationUnitConverter.DurationSerialization.deserializer;
+
+
```
cowtowncoder, line 13: Could you add `@since 2.12` here?

cowtowncoder:
LGTM! @kupci WDYT?

kupci:
Yes, just looking at this in fact, and (much as I tried), could not find any issues, so yes, LGTM. Maybe @beamerblvd ?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/DurationSerializer.java
```diff
@@ -135,4 +183,9 @@ protected JsonToken serializationShape(SerializerProvider provider) {
     protected JSR310FormattedSerializerBase<?> withFeatures(Boolean writeZoneId, Boolean writeNanoseconds) {
         return new DurationSerializer(this, _useTimestamp, writeNanoseconds, _formatter);
     }
+
+    @Override
+    protected DateTimeFormatter _useDateTimeFormatter(SerializerProvider prov, JsonFormat.Value format) {
+        return null;
```
kupci, line 189: This is the part I was stuck on a bit, suddenly thinking the DateTimeFormatter could be put to use, and then after spinning on that a bit, found I agree with whoever added this comment (line 103 of original `DurationSerializer`.
 // Does not look like we can make any use of DateTimeFormatter here?

kupci (approved):

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/util/DurationUnitConverter.java
obarcelonap: Done, I added documentation to the class too.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/DurationSerializer.java
obarcelonap: This is because serializer is extending `JSR310FormattedSerializerBase` which has a common `DateTimeFormatter`. But `Duration` is not supported by `DateTimeFormatter`, actually it has its own formatter embedded in the class based on ISO-8601. I removed the comment because once this is clear, does not make sense to use the formatter of the base class in the `DurationSerializer`.

For the long term, I think `JSR310FormattedSerializerBase` can be refactored a bit extracting the common functionality of classes extending `DateTime` into its own base (maybe `JSR310FormattedSerializerDateTimeBase`), so there is no need to override `_useDateTimeFormatter` in this case.

Hope this clarifies a bit my point of view.

cowtowncoder:
Ok, will merge now!

PR #195, opened 2020-11-19
Fix OptionalDouble deserialization of special values without coercion

carterkozak:
floating point special values as JSON Strings are considered "native"
because they cannot be represented as JSON numbers.

See:
https://github.com/FasterXML/jackson-databind/pull/2942
https://github.com/FasterXML/jackson-databind/commit/7705d35fc0f58e5aec389b2cdadf9a046a55b00b


cowtowncoder:
Good catch, thank you!

cowtowncoder:
Ah. One complication: in 3.0, these are merged in `jackson-databind`. Could you also do another PR against master?

carterkozak:
Of course :-)

PR #197, opened 2021-01-07
Fix #196 ZonedDateTime deserialization skipped @JsonFormat features o…

### LINKED ISSUE #196: `@JsonFormat` overriden features don't apply when there are no other options while deserializing ZonedDateTime
After deserializing a following class:
```
    static class WrapperWithFeatures {
        @JsonFormat(without = JsonFormat.Feature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE)
        public ZonedDateTime value;

        public WrapperWithFeatures() { }
        public WrapperWithFeatures(ZonedDateTime v) { value = v; }
    }
```
`value` field should preserve timezone from the source - but it doesn't. Following test fails:
```
    @Test
    public void testDeserializationWithZonePreserved() throws Throwable
    {
        WrapperWithFeatures wrapper = JsonMapper.builder()
                .build()
                .readerFor(WrapperWithFeatures.class)
                .readValue("{\"value\":\"2000-01-01T12:00+01:00\"}");
        assertEquals("Timezone should be preserved.",
                ZonedDateTime.of(2000, 1, 1, 12, 0, 0 ,0, ZoneOffset.ofHours(1)),
                wrapper.value);
    }
```

MaciejDobrowolski:
…verride if there were no other options defined on it

### DIFF
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ModuleTestBase.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ModuleTestBase.java
@@ -2,6 +2,7 @@ package com.fasterxml.jackson.datatype.jsr310;
 
 import java.time.ZoneId;
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.TimeZone;
 import java.util.Collections;
 import java.util.Map;
@@ -28,7 +29,8 @@ public class ModuleTestBase
 
     protected static MapperBuilder<?,?> newMapperBuilder() {
         return JsonMapper.builder()
-                .addModule(new JavaTimeModule());
+                .addModule(new JavaTimeModule())
+                .defaultLocale(Locale.ENGLISH);
     }
 
     protected static MapperBuilder<?,?> newMapperBuilder(TimeZone tz) {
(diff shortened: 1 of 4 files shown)


datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ModuleTestBase.java
```diff
@@ -28,7 +29,8 @@ protected static ObjectMapper newMapper() {
 
     protected static MapperBuilder<?,?> newMapperBuilder() {
         return JsonMapper.builder()
-                .addModule(new JavaTimeModule());
+                .addModule(new JavaTimeModule())
+                .defaultLocale(Locale.ENGLISH);
```
MaciejDobrowolski, line 33: the tests (parsing dates like `1 Jan 2020`) were failing when you ran them on a pc with a locale different than English

cowtowncoder:
First of all: thank you for both reporting the issue and especially offering a PR! (+1 for test too).

Quick request: since `master` is future Jackson 3.0, not to be released very soon, could you rebase this against 2.12?

Also: unless I have asked for an CLA (apologies if I did in the past), I'd need it:

before merging the PR -- only needs to be done once, before the first contribution, good thereafter.

The common way is to print pdf, fill & sign, scan/take photo, email to `info` at fasterxml dot com.

Looking forward to reviewing and merging the patch!

MaciejDobrowolski:
Hey, just created another PR for 2.12, it was just easier this way :) Closing this one

PR #198, opened 2021-01-08
Fix #196 ZonedDateTime deserialization skipped @JsonFormat features o…

### LINKED ISSUE #196: `@JsonFormat` overriden features don't apply when there are no other options while deserializing ZonedDateTime

MaciejDobrowolski:
…verride if there were no other options defined on it. Added a fixed English `Locale` to tests b/c they were not passing in non-English environment.

**CLA has been signed**

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -187,15 +187,13 @@ public class InstantDeserializer<T extends Temporal>
     {
         InstantDeserializer<T> deserializer =
                 (InstantDeserializer<T>)super.createContextual(ctxt, property);
-        if (deserializer != this) {
-            JsonFormat.Value val = findFormatOverrides(ctxt, property, handledType());
-            if (val != null) {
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Ok this makes sense. I am not sure why the guard statement was like it was but I can see why that prevented correct working.
I think I'll make some slight changes beyond this PR, but will get it merged first. Thanks!

Issue #199, opened 2021-01-18
In 2.11.4 set [pattern = "yyyy-MM-dd"] in @JsonFormat cannot deserialize a valid LocalDate.


Fatezhang:
__Hi there, I'm using SpringBoot 2.4.2 which Jackson version is 2.11.4, got an issue:__

When I set the `lenient` to false, the string `2021-01-18` cannot be deserialized to a LocalDate, see:

```java
    @JsonFormat(pattern = "yyyy-MM-dd", lenient = OptBoolean.FALSE)
    private LocalDate localDate;
```
and then I send the data in request body like:
```shell
curl -L -X GET 'http://localhost:8080/test' \
-H 'Content-Type: application/json' \
--data-raw '{
    "localDate": "2020-01-18"
}'
```
It shows
```
... could not be parsed: Unable to obtain LocalDate from TemporalAccessor: ...
```
But when I set it to `TRUE` or remove the annotation, it works.
I think set `lenient` to false means using strict mode, but `2020-01-18` should be a valid LocalDate, am I wrong?

---

Found an issue but not sure whether it related:
 #148 

---

## Update

`@JsonFormat(pattern = "yyyy-MM-dd", lenient = OptBoolean.FALSE)` cannot work,

__but__

`@JsonFormat(lenient = OptBoolean.FALSE)` works.

Could someone help to explain this behaviour? 
The above issue is in version `2.11.4`(or 2.11), but when I downgrade the version to `2.10.2`, `@JsonFormat(pattern = "yyyy-MM-dd", lenient = OptBoolean.FALSE)` works when I pass  `{"localDate": "2020-01-18"}`.

cowtowncoder:
Since there is a difference between versions, it sounds like there might be a problem. If possible, it'd be good to see if 2.12.1 works or not; but most importantly it'd be great to have an actual unit test based on description above.

Fatezhang:
Ah....... I found the actual reason.

The pattern should be `uuuu-MM-dd`.

Java 8 uses 'uuuu' for year, not 'yyyy' . In Java 8, 'yyyy' means “year of era” (BC or AD). See [here](https://howtodoinjava.com/java/date-time/resolverstyle-strict-date-parsing/).

[...]

PR #201, opened 2021-01-20
Add/update unit tests when allow strict custom format

Fatezhang:
## WHY

See issue #199 

We'd better add some unit tests to make sure the original behaviour works when we allow strict `LocalDate` parsing. 

## HOW

- Add test cases for format pattern `yyyy-MM-dd G` in strict mode(`y` means `year of the era` in `DateTimeFormatter` and should use together with `G`. ).
- Add test cases for format pattern `uuuu-MM-dd` in strict mode.

### DIFF
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
@@ -65,6 +65,24 @@ public class LocalDateDeserTest extends ModuleTestBase
         public StrictWrapper(LocalDate v) { value = v; }
     }
 
+    final static class StrictWrapperWithYearOfEra {
+        @JsonFormat(pattern="yyyy-MM-dd G",
+                lenient = OptBoolean.FALSE)
+        public LocalDate value;
+
+        public StrictWrapperWithYearOfEra() { }
+        public StrictWrapperWithYearOfEra(LocalDate v) { value = v; }
+    }
+
+    final static class StrictWrapperWithYearWithoutEra {
+        @JsonFormat(pattern="uuuu-MM-dd",
+                lenient = OptBoolean.FALSE)
+        public LocalDate value;
+
+        public StrictWrapperWithYearWithoutEra() { }
+        public StrictWrapperWithYearWithoutEra(LocalDate v) { value = v; }
+    }
+
     /*
     /**********************************************************
     /* Deserialization from Int array representation
@@ -315,9 +333,49 @@ public class LocalDateDeserTest extends ModuleTestBase
 
     // for [modules-java8#148]
     @Test(expected = InvalidFormatException.class)
-    public void testStrictCustomFormat() throws Exception
+    public void testStrictCustomFormatForInvalidFormat() throws Exception
     {
-        /*StrictWrapper w =*/ MAPPER.readValue("{\"value\":\"2019-11-31\"}", StrictWrapper.class);
+        /*StrictWrapper w =*/ MAPPER.readValue("{\"value\":\"2019-11-30\"}", StrictWrapper.class);
+    }
+
+    @Test(expected = InvalidFormatException.class)
+    public void testStrictCustomFormatForInvalidFormatWithEra() throws Exception
+    {
+        /*StrictWrapperWithYearOfEra w =*/ MAPPER.readValue("{\"value\":\"2019-11-30\"}", StrictWrapperWithYearOfEra.class);
+    }
+
+    @Test(expected = InvalidFormatException.class)
+    public void testStrictCustomFormatForInvalidDateWithEra() throws Exception
+    {
+        /*StrictWrapperWithYearOfEra w =*/ MAPPER.readValue("{\"value\":\"2019-11-31 AD\"}", StrictWrapperWithYearOfEra.class);
+    }
+
+    @Test
+    public void testStrictCustomFormatForValidDateWithEra() throws Exception
+    {
+        StrictWrapperWithYearOfEra w = MAPPER.readValue("{\"value\":\"2019-11-30 AD\"}", StrictWrapperWithYearOfEra.class);
+
+        assertEquals(w.value, LocalDate.of(2019, 11, 30));
+    }
+
+    @Test(expected = InvalidFormatException.class)
+    public void testStrictCustomFormatForInvalidFormatWithoutEra() throws Exception
+    {
+        /*StrictWrapperWithYearWithoutEra w =*/ MAPPER.readValue("{\"value\":\"2019-11-30 AD\"}", StrictWrapperWithYearWithoutEra.class);
+    }
+
+    @Test(expected = InvalidFormatException.class)
+    public void testStrictCustomFormatForInvalidDateWithoutEra() throws Exception
+    {
+        /*StrictWrapperWithYearWithoutEra w =*/ MAPPER.readValue("{\"value\":\"2019-11-31\"}", StrictWrapperWithYearWithoutEra.class);
+    }
+
+    @Test
+    public void testStrictCustomFormatForValidDateWithoutEra() throws Exception
+    {
+        StrictWrapperWithYearWithoutEra w = MAPPER.readValue("{\"value\":\"2019-11-30\"}", StrictWrapperWithYearWithoutEra.class);
+
+        assertEquals(w.value, LocalDate.of(2019, 11, 30));
     }
 
     /*
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserTest.java
@@ -68,6 +68,24 @@ public class LocalDateTimeDeserTest
         public StrictWrapper(LocalDateTime v) { value = v; }
     }
 
+    final static class StrictWrapperWithYearOfEra {
+        @JsonFormat(pattern="yyyy-MM-dd HH:mm G",
+                lenient = OptBoolean.FALSE)
+        public LocalDateTime value;
+
+        public StrictWrapperWithYearOfEra() { }
+        public StrictWrapperWithYearOfEra(LocalDateTime v) { value = v; }
+    }
+
+    final static class StrictWrapperWithYearWithoutEra {
+        @JsonFormat(pattern="uuuu-MM-dd HH:mm",
+                lenient = OptBoolean.FALSE)
+        public LocalDateTime value;
+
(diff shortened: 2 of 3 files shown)


kupci:
This looks like a good addition, but can the 'uuuu' format be added as a format, instead of replacing the 'yyyy'? Also, it seems your 'use case' from the original bug would be suitable test case here, to show the difference between the 'year-of-era' and 'year' formats: 
```
DateTimeFormatter formatter = DateTimeFormatter
    .ofPattern("yyyy-MM-dd G")
    .withResolverStyle(ResolverStyle.STRICT);

LocalDate newDate = LocalDate.parse("2021-01-20 AD", formatter);
 
```

cowtowncoder:
Quick question: could this be rebased against 2.12 or 2.13? `master` is still for upcoming Jackson 3.0.0 which will not be released soon.

Fatezhang:
> Quick question: could this be rebased against 2.12 or 2.13? `master` is still for upcoming Jackson 3.0.0 which will not be released soon.

Sure, will rebase to 2.13. 👍

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
```diff
@@ -315,9 +324,31 @@ public void testCustomFormat() throws Exception
...
-        /*StrictWrapper w =*/ MAPPER.readValue("{\"value\":\"2019-11-31\"}", StrictWrapper.class);
+        StrictWrapper w = MAPPER.readValue("{\"value\":\"2019-11-30 AD\"}", StrictWrapper.class);
+
+        assertEquals(w.value, LocalDate.of(2019, 11, 30));
+    }
+
+    @Test(expected = InvalidFormatException.class)
+    public void testStrictCustomFormatWithInvalidDateWithoutEra() throws Exception
+    {
+        /*StrictWrapper w =*/ MAPPER.readValue("{\"value\":\"2019-11-31\"}", StrictWrapperWithYearWithoutEra.class);
```
LangInteger, line 343: The comment should be /*StrictWrapperWithYearWithoutEra w=*/

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserTest.java
```diff
@@ -513,19 +522,42 @@ public void testDeserializationCaseInsensitiveDisabled_InvalidDate() throws Thro
...
+    {
+        StrictWrapper w = MAPPER.readValue("{\"value\":\"2019-11-30 20:45 AD\"}", StrictWrapper.class);
+
+        assertEquals(w.value, LocalDateTime.of(2019, 11, 30, 20, 45));
+    }
+
+    @Test(expected = InvalidFormatException.class)
+    public void testStrictCustomFormatInvalidTimeWithoutEra() throws Exception
+    {
+        /*StrictWrapper w =*/ MAPPER.readValue("{\"value\":\"2019-11-31 25:45\"}", StrictWrapperWithYearWithoutEra.class);
```
LangInteger, line 551: The comment should be `/*StrictWrapperWithYearWithoutEra w =*/`

kupci (approved):

kupci:
I think this is close to getting ready. One thing you'll (if not sent already) @cowtowncoder will need is the CLA, from:

and usually the easiest way is to print, fill & sign, scan/photo, email to info at fasterxml dot com.

Thank you once again for the PR!

cowtowncoder:
I am not sure it is good to only change existing test to add era (usage that probably most users don't do), instead of adding separate tests for that case? That is, wouldn't it be better to have both existing ones and new ones added?

kupci:
Yes, I'll add a comment to the PR - we should keep the original tests as-is, since this is the typical use case, and then also add the 'date with era' as it's own new test.

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
```diff
@@ -57,20 +57,29 @@ public ShapeWrapper() { }
     }
 
     final static class StrictWrapper {
-        @JsonFormat(pattern="yyyy-MM-dd",
+        @JsonFormat(pattern="yyyy-MM-dd G",
```
kupci, line 60: Instead of changing this test format, we should keep this one 'as-is', and then add entirely new tests (as you have done) for the 'date-with-era' test case.

Based on this SO post, most users will probably default to using 'yyyy', so there should be test coverage for that scenario. Thus, recommend having three Wrapper classes:

 @JsonFormat(pattern="yyyy-MM-dd" // original, leaving as-is (year-of-era, but lacks era)

 @JsonFormat(pattern="yyyy-MM-dd G" // new, improved 'year-of-era', with era

@JsonFormat(pattern="uuuu-MM-dd", // new 'year'

https://stackoverflow.com/questions/41177442/uuuu-versus-yyyy-in-datetimeformatter-formatting-pattern-codes-in-java

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserTest.java
```diff
@@ -60,14 +60,23 @@
     private final TypeReference<Map<String, LocalDateTime>> MAP_TYPE_REF = new TypeReference<Map<String, LocalDateTime>>() { };
 
     final static class StrictWrapper {
-        @JsonFormat(pattern="yyyy-MM-dd HH:mm",
+        @JsonFormat(pattern="yyyy-MM-dd HH:mm G",
```
kupci, line 63: Same comment as previously, this format should stay 'as-is' (without the era), so we still test the original test case (which is the common scenario), and then have new test cases for the 'year-with-era' scenario ('yyyy-MM-dd HH:mm G') and the 'year' scenario ('uuuu-MM-dd HH:mm')

kupci (changes requested):

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
kupci: I am a little unsure about the 'yyyy-MM-dd G' format after reading the StackOverflow post. Maybe we can trim this down to just 'yyyy-MM-dd' (era ambiguous, but typical usage) and 'uuuu-MM-dd' (no ambiguity). 

> We should also see that using eras (symbol "G") in context of ISO is in general dangerous if we consider historic dates. If "G" is used with "u" then both fields are unrelated to each other. And if "G" is used with "y" then the formatter is satisfied but still uses proleptic gregorian calendar when the historic date mandates different calendars and date-handling.

However, in another answer in the same post, they make this point, so probably useful to have both:

> Conclusion: If you use y you should also use G. Since G is rarely used, the correct year symbol is u, not y, otherwise a non-positive year will show incorrectly.

The problems seem to be due to the fact that the designers of [JSR-310](https://jcp.org/en/jsr/detail?id=310) decided to use the [CLDR ](https://en.wikipedia.org/wiki/Common_Locale_Data_Repository)conventions, but:

> .. this decision to follow CLDR was not quite consistent because JSR-310 had also introduced new pattern symbols which didn't and still don't exist in CLDR

Fatezhang: Yes, thanks @kupci for your reply, I agree with you to have three Wrapper classes you mentioned above and leave the original test case there. 
But for the `yyyy-MM-dd`, I assume the original test case want to test an invalid date("2019-11-31" is invalid) rather an invalid format:
```java
@Test(expected = InvalidFormatException.class)
public void testStrictCustomFormat() throws Exception
{
    /*StrictWrapper w =*/ MAPPER.readValue("{\"value\":\"2019-11-31\"}", StrictWrapper.class);
}
```
For any valid date in this test case is invalid format. So would you mind me to change the `2019-11-31` to `2019-11-30` and cover the `invalid date case` in another test case?

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserTest.java
Fatezhang: Same as the above, the pattern `yyyy-MM-dd HH:mm` doesn't allow any valid date-time(like `2019-11-30 15:45`) in strict mode, would you mind me update the original test case to say `test the invalid format` rather `test the invalid date time`. And cover the `invalid date time` in other test cases.

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
kupci: Yes, I see your point now, the original test didn't actually test what it was intended to test, or should have tested.

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserTest.java
kupci: Yes, sounds good.

kupci (approved):
Looks good to me. Thanks for adding the extra test scenarios, some of those I hadn't even thought of. My feeling is that others (myself included) will gradually move to the 'uuuu' as it gains more widespread acceptance, so good to have extra testing around this. We'll look for @cowtowncoder 's comments.

cowtowncoder:
Looks good to me too, got CLA and will now merge.

cowtowncoder:
Ok, so... does anyone want to propose either changing default year format to use `uuuu` or -- possibly better -- add configurability to this module to allow changing default with a feature?
This would only affect cases where defaults are not overridden by annotation or config overrides for types.

kupci:
Figuring this would be a good candidate for the [date-time-config](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-5) work? Though I have a question about the point below:

_This would only affect cases where defaults are not overridden by annotation or config overrides for types._

So this wouldn't apply in this particular example (#199 ) , since the default can be overridden by annotation, however, we would support this for other cases where there isn't an annotation / config override?

cowtowncoder:
Correct, it would not apply if `pattern` of `@JsonFormat` is specified (assuming we can distinguish that case... I think we can), would only be used for selecting the default pattern.

For `DateTimeConfig` we could also specify tri-state operation (true / false / default), which would allow explicit yyyy vs uuuu override, if that matters.

Still, not sure if this functionality is actually requested or not I guess.
If there is desire to do this, we'd need a separate issue.

Issue #202, opened 2021-01-20
Unable to deserialize `YearMonth` when running as java9 module, added with `@JsonDeserialize` annotation


walkeros:
Consider following field annotated with YearMonth deserializer:
```

class MyClass {
  @JsonSerialize(using = YearMonthSerializer.class)
  @JsonDeserialize(using = YearMonthDeserializer.class)
  private final YearMonth period;
}
```

It looks to impossible to deserialize in such a way when running your app as named (java9) module. You will get:

```
java.lang.reflect.InaccessibleObjectException: Unable to make private com.fasterxml.jackson.datatype.jsr310.deser.YearMonthDeserializer() accessible: module com.fasterxml.jackson.datatype.jsr310 does not "opens com.fasterxml.jackson.datatype.jsr310.deser" to module com.fasterxml.jackson.databind
        at java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:340) ~[?:?]
        at java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:280) ~[?:?]
        at java.lang.reflect.Constructor.checkCanSetAccessible(Constructor.java:189) ~[?:?]
        at java.lang.reflect.Constructor.setAccessible(Constructor.java:182) ~[?:?]
        at com.fasterxml.jackson.databind.util.ClassUtil.checkAndFixAccess(ClassUtil.java:934) ~[jackson-databind-2.11.4.jar:?]
        at com.fasterxml.jackson.databind.util.ClassUtil.findConstructor(ClassUtil.java:568) ~[jackson-databind-2.11.4.jar:?]
        at com.fasterxml.jackson.databind.util.ClassUtil.createInstance(ClassUtil.java:550) ~[jackson-databind-2.11.4.jar:?]
        at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.deserializerInstance(DefaultDeserializationContext.java:229) ~[jackson-databind-2.11.4.jar:?]
        at com.fasterxml.jackson.databind.deser.BasicDeserializerFactory.findDeserializerFromAnnotation(BasicDeserializerFactory.java:2099) ~[jackson-databind-2.11.4.jar:?]
        at com.fasterxml.jackson.databind.deser.BasicDeserializerFactory.constructCreatorProperty(BasicDeserializerFactory.java:1019) ~[jackson-databind-2.11.4.jar:?]
[...]

PR #203, opened 2021-01-21
Updated the access of YearMonthDeserialiser constructor from protected

shubhendu-goswami:
## WHY
Issue - [https://github.com/FasterXML/jackson-modules-java8/issues/202](#202)

## HOW
Updated the access modifier of YearMonthDeserialiser constructor from protected to public.


cowtowncoder:
Thanks! I think I'll want this against 2.12 branch, so will make the change by hand. Thanks!

Issue #131, opened 2019-08-23
Deserializing ZonedDateTime with basic TZ offset notation (0000)


noway1979:
I stumbled upon deserialization issues when the given string equals to ISO8601 "basic" format. 

`2019-08-22T12:36:46.361+0000`

Exception message: 
```
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.ZonedDateTime` from String "2019-08-22T12:36:46.361+0000": Failed to deserialize java.time.ZonedDateTime: (java.time.format.DateTimeParseException) Text '2019-08-22T12:36:46.361+0000' could not be parsed at index 23
 at [Source: (String)""2019-08-22T12:36:46.361+0000""; line: 1, column: 1]
```
This format misses out a colon in offset in comparison to extended format (as described at: https://bugs.openjdk.java.net/browse/JDK-8176547). 

`2019-08-22T12:36:46.361+00:00`

The issue states that java.time.format.DateTimeFormatter.ISO_ZONED_DATE_TIME does not support this basic format.

However, as this format is provided by Spring exception handler I was wondering if there exists any workaround to successfully parse this format. Maybe I am simply missing the obvious here?

As current workaround I had to register a custom InstantDeserializer and set com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.replaceZeroOffsetAsZ to true. 
Any hint is very much appreciated.

## Reproduce
```
		ZonedDateTime zdt1 = ZonedDateTime.now();
		ObjectMapper mapper = new ObjectMapper()                                         
                                        .registerModule(new ParameterNamesModule())
					.registerModule(new Jdk8Module())
                                        .registerModule(new JavaTimeModule());

		//happy path: serialize and deserialize
		String zdt1S = mapper.writeValueAsString(zdt1);
		ZonedDateTime zdt1R = mapper.readValue(zdt1S, ZonedDateTime.class);
		assertThat(zdt1R, notNullValue());
		gLogger.info("zdt1S: " + zdt1R);

		//ZonedDateTime with basic formatted offset leads to exception
		String basicTS = "\"2019-08-22T12:36:46.361+0000\"";
		ZonedDateTime zdt2 = mapper.readValue(basicTS, ZonedDateTime.class);
		gLogger.info("zdt2S: " + zdt2);
```

kupci:
> However, as this format is provided by Spring exception handler 

[...]

PR #208, opened 2021-03-21
Added a regex checker and fixer for offsets without colons for ZDateTime and OSDateTime

oeystein:
## WHY
See issue #131 for details 

Colonless offsets for ISO 8601s strings a common way of representing DateTimes with offsets. Unfortunately ZonedDateTime.parse(string) and OffsetDateTime.parse(string) does not support this, breaking jackson compatibility. 

## HOW
Uses a regex to check for offsets without colons, and if found - adds in a colon to make it compatible with ZonedDateTime.parse() and OffsetDateTime.parse()

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -65,7 +65,7 @@ public class InstantDeserializer<T extends Temporal>
      *
      * @since 2.13
      */
-    private static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("[+-][0-9]{4}");
+    private static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("(?<!^)[+-][0-9]{4}");
 
     public static final InstantDeserializer<Instant> INSTANT = new InstantDeserializer<>(
             Instant.class, DateTimeFormatter.ISO_INSTANT,
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -65,7 +65,7 @@ public class InstantDeserializer<T extends Temporal>
      *
      * @since 2.13
      */
-    protected static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("(?<!^)[+-][0-9]{4}");
+    protected static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("[+-][0-9]{4}(?=\\[|$)");
 
     public static final InstantDeserializer<Instant> INSTANT = new InstantDeserializer<>(
             Instant.class, DateTimeFormatter.ISO_INSTANT,
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -38,6 +38,7 @@ import java.time.temporal.TemporalAccessor;
 import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
@@ -58,6 +59,13 @@ public class InstantDeserializer<T extends Temporal>
      * @since 2.9.0
      */
     private static final Pattern ISO8601_UTC_ZERO_OFFSET_SUFFIX_REGEX = Pattern.compile("\\+00:?(00)?$");
+
+    /**
+     * Constants used to check if ISO 8601 time string is colonless. See [jackson-modules-java8#131]
+     *
+     * @since 2.13
+     */
+    private static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("[+-][0-9]{4}");
 
     public static final InstantDeserializer<Instant> INSTANT = new InstantDeserializer<>(
             Instant.class, DateTimeFormatter.ISO_INSTANT,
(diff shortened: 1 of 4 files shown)


oeystein:
Looks like it breaks negative time strings (i was not actually aware this was a thing ... ) Need to solidify the regex expression.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -65,7 +65,7 @@
      *
      * @since 2.13
      */
-    private static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("[+-][0-9]{4}");
+    private static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("(?<!^)[+-][0-9]{4}");
```
kupci, line 68: Should be some tests for the sign handling?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -277,6 +285,16 @@ protected T _fromString(JsonParser p, DeserializationContext ctxt,
             string = replaceZeroOffsetAsZIfNecessary(string);
         }
 
+        // For some reason DateTimeFormatter.ISO_INSTANT only supports UTC ISO 8601 strings, so it have to be excluded
+        if (_formatter == DateTimeFormatter.ISO_OFFSET_DATE_TIME ||
+            _formatter == DateTimeFormatter.ISO_ZONED_DATE_TIME) {
+
+            // 21-March-2021, Oeystein: Work-around for supporting iso-8601 time offsets that are
```
kupci, line 292: I'd suggest mentioning that this is to support **basic** 8601 format (colonless) vs. **extended** format, to connect the JavaDoc to the 8601 specification.

kupci (commented):
Just a few minor points otherwise looks good to me.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
oeystein: Added some additional tests, are they sufficient or did you have something else in mind?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
oeystein: Reworded the comment.

kupci (approved):
Looks good to me, @cowtowncoder thoughts?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
kupci: Ok, I did think of one more thing...for the regular exp., why not restrict this to the last 5 characters of the string? Two reasons this might be nice (1) limits the possibility that there is some odd combination such that there are 4 digits in the middle of the string, that's something else besides the offset, and we don't want a colon inserted there (2) performance gain - admittedly very small, but regexs can easily get out of hand. This one here is perfectly fine, but for some interesting reading on the subject: https://regular-expressions.mobi/catastrophic.html?wlr=1

kupci:
FYI: I naively thought it would be better to apply the regex to the end of the string, and quickly found out: not so simple. I think given the shortness of the string, and the simplicity (clean and simple) of the regex you have, it should be fine, so please disregard (now deleted) comment I had about the regex.

oeystein:
Well that explains why the comment button refused to work when I tried to reply!  

And originally i had the same thought as you. I overdid the regex expression before figuring out simple ought to do the trick. There are also variations of the ISO 8601 string which includes TZ on end as mentioned in #131, such as "2000-01-01T12:00+0100[Europe/Paris]". So any at the end of expression would have to check for it to be the end or followed by "[".

kupci:
@oeystein Thanks! Good points, the regex can get crazy quick.

I think this is getting ready, though @cowtowncoder might have some further comments, in the meantime, one thing you'll need to send in (if not already) the CLA, from:

and usually the easiest way is to print, fill & sign, scan/photo, email to info at fasterxml dot com.

Thank you once again for the PR!

oeystein:
@kupci Any time! 

And i just sent over the contributor agreement so the formalities should be in order now.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -59,6 +60,13 @@
      */
     private static final Pattern ISO8601_UTC_ZERO_OFFSET_SUFFIX_REGEX = Pattern.compile("\\+00:?(00)?$");
 
+    /**
+     * Constants used to check if ISO 8601 time string is colonless. See [jackson-modules-java8#131]
+     *
+     * @since 2.13
+     */
+    protected static final Pattern ISO8601_COLONLESS_OFFSET_REGEX = Pattern.compile("(?<!^)[+-][0-9]{4}");
```
cowtowncoder, line 68: Quick question here: I understand that first part is to try to avoid false matches (since we cannot assume it's at the end, due to possible timezone suffix), but what does the part in parenthesis mean? :)

Also wonder if it could/should check trailing word boundary (either end-of-string, or `[`) -- that would anchor it correctly, I think.
Just curious, not sure if there are performance implications; I know some matching constructors can be more expensive than others.

cowtowncoder:
@oeystein Thank you for contributing this (including tests!) & sending CLA -- only have one question, hoping to merge this soon.

oeystein:
Re-posting this as a pull request comment rather then a code reply comment since it's been hanging unapproved for a couple days now.  

----------------

Been giving the regex expression some thought over the last few days and i agree its not optimal. The parenthesis just simply uses anchors to ignore checking the first letter in the the string.

We can however likewise simply use match on anchor(end of line) or boundary([) to check for line ends or "[" matches. So realistically i don't believe the following example should not have any major performance implications? It's at least the best i appear to come up with, and the one i believe would be the best.

[+-][0-9]{4}(?=[|$)

Another solution is too only look for matches after T(which is pretty much the only guaranteed thing in ISO 8601 DateTime formats), which should also technically work but not created a regex expression for this yet.

@kupci I believe you also looked a bit into the performance complications of this, any comments?

oeystein:
Committed the requested changes, so if everything seems good now this should be ready to be merged.

cowtowncoder:
Sorry, extremely heavy week. This is near the top of the pile (https://github.com/FasterXML/jackson-future-ideas/wiki/Jackson-Work-in-Progress) so while delayed, not forgotten, and I intend to get to it this week, hopefully tomorrow.
Thank you for your patience @oeystein.

PR #211, opened 2021-03-26
Simplify handling of `LocalDate` value with invalid trailing time-portion

cowtowncoder:
As per title: test change just cleaning up, but for `LocalDate`, avoid unnecessary handling of timezone.


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserializer.java
```diff
@@ -155,7 +155,8 @@ protected LocalDate _fromString(JsonParser p, DeserializationContext ctxt,
                 // JavaScript by default includes time in JSON serialized Dates (UTC/ISO instant format).
                 if (string.length() > 10 && string.charAt(10) == 'T') {
                    if (string.endsWith("Z")) {
-                       return LocalDateTime.ofInstant(Instant.parse(string), ZoneOffset.UTC).toLocalDate();
+                       return LocalDate.parse(string.substring(0, string.length() - 1),
+                               DateTimeFormatter.ISO_LOCAL_DATE_TIME);
```
cowtowncoder, line 159: Seems to me that there is no need to go back and forth to zoned instant when simply trying to retain local value.
Tests all pass as well.

cowtowncoder:
@kupci Just noticed this when looking at LocalDate/LocalDateTime deserialization; wanted to sanity check my thinking here.

kupci (approved):
> Should not contain time component when 'strict' mode set for property or type (enable 'lenient' handling to allow)

Nice solution! I see this question coming up every so often, so this looks like a good improvement.

cowtowncoder:
I copied that from `LocalDateTime` deserializer where it already existed. I hope logic makes sense; it should as lenient is the default.

I actually should file an issue for that change, and simplification can just be part of implementation for that issue.

Issue #221, opened 2021-07-01
Serialization of ZonedDateTime with ZoneId bug since 2.12


vmeunier:
Hello,

I have a simple test case, where I expect my ObjectMapper to keep the ZoneId I've put in my code, for example UTC, Europe/Paris, ... during Serialization.

If I read the documentation

For TimeZone handling, ADJUST_DATES_TO_CONTEXT_TIME_ZONE (default: true) specifies whether the context provided by java.time.TimeZone 'SerializedProvider#getTimeZone()' should be used to adjust Date/Time values on deserialization, even if the value itself contains timezone information. If the value is OffsetDateTime.MIN or OffsetDateTime.MAX, the Date/Time value will not be adjusted. **If disabled, it will only be used if the value itself does not contain any TimeZone information.**

Here is my ObjectMapper configuration : 

```java
        final JavaTimeModule module = new JavaTimeModule();
        mapper = new ObjectMapper() //
            .setSerializationInclusion(Include.NON_NULL) //
            .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS) //
            .disable(DeserializationFeature.ADJUST_DATES_TO_CONTEXT_TIME_ZONE) //
            .setTimeZone(TimeZone.getTimeZone("Europe/Paris")) //
            .registerModule(module);
```

If I test : 

```
        final BeanForJson bean = new BeanForJson();
        bean.setZonedDateTime(ZonedDateTime.of(2016, 10, 15, 17, 51, 50, 0, ZoneId.of("UTC")));

        final String json = mapper.writeValueAsString(bean);
        assertThat(json).isEqualTo("{\"zonedDateTime\":\"2016-10-15T17:51:50Z\"}");
```

I get : 

```
org.opentest4j.AssertionFailedError: 
expected: "{"zonedDateTime":"2016-10-15T17:51:50Z"}"
but was : "{"zonedDateTime":"2016-10-15T19:51:50+02:00"}"
```

It is working in 2.11.4 (same use-case).

Thanks

kupci:
[...]

PR #222, opened 2021-07-09
Serialization : adjust the TimeZone of dates if WRITE_DATES_WITH_CONTEXT_TIME_ZONE is specified

vmeunier:
#221

### DIFF
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
@@ -273,35 +273,62 @@ public class ZonedDateTimeSerTest
         assertEquals("The value is incorrect.", "\"" + DateTimeFormatter.ISO_ZONED_DATE_TIME.format(date) + "\"", value);
     }
 
+    @Test
+    public void testSerializationAsStringWithDefaultTimeZoneAndContextTimeZoneOnAndACustomFormatter() throws Exception {
+        ZonedDateTime date = ZonedDateTime.now(Z3);
+        // With a custom DateTimeFormatter without a ZoneId.
+        String value = MAPPER.registerModule(new SimpleModule().addSerializer(new ZonedDateTimeSerializer(FORMATTER_WITHOUT_ZONEID))).writer()
+                .with(TimeZone.getTimeZone(Z2))
+                .without(SerializationFeature.WRITE_DATES_WITH_ZONE_ID)
+                .without(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+                .with(SerializationFeature.WRITE_DATES_WITH_CONTEXT_TIME_ZONE)
+                .writeValueAsString(date);
+
+        // We expect to have the date written with the ZoneId Z2
+        assertEquals("The value is incorrect", "\"" + date.withZoneSameInstant(Z2).format(FORMATTER_WITHOUT_ZONEID) + "\"", value);
+    }
+
+    @Test
+    public void testSerializationAsStringWithDefaultTimeZoneAndContextTimeZoneOffAndACustomFormatter() throws Exception {
+        ZonedDateTime date = ZonedDateTime.now(Z3);
+        // With a custom DateTimeFormatter without a Zone.
+        String value = MAPPER.registerModule(new SimpleModule().addSerializer(new ZonedDateTimeSerializer(FORMATTER_WITHOUT_ZONEID))).writer()
+                .with(TimeZone.getTimeZone(Z2))
+                .without(SerializationFeature.WRITE_DATES_WITH_ZONE_ID)
+                .without(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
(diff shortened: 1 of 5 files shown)


vmeunier:
@kupci @cowtowncoder 

Tell me if you think I should add more tests, or if anything is wrong at all :)

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
```diff
@@ -270,6 +273,37 @@ public void testSerializationAsStringWithZoneIdOn() throws Exception {
         assertEquals("The value is incorrect.", "\"" + DateTimeFormatter.ISO_ZONED_DATE_TIME.format(date) + "\"", value);
     }
 
+
+    @Test
+    public void testSerializationAsStringWithDefaultTimeZoneAndContextTimeZoneOn() throws Exception {
+		ZonedDateTime date = ZonedDateTime.now(Z3);
+		// With a custom DateTimeFormatter without a ZoneId.
+		String value = MAPPER.registerModule(new SimpleModule().addSerializer(new ZonedDateTimeSerializer(FORMATTER_WITHOUT_ZONEID))).writer()
+		        .with(TimeZone.getTimeZone(Z2))
```
kupci, line 282: Any other tests needed for the various `SerializationFeature` combinations? Speaking of eclipse, it would be a useful exercise to run a code coverage (JaCoCo or similar) to see if any paths are missed.

kupci: Taking a 2nd look, I think you have enough test coverage for `ZonedDateTimeSerTest`, given it really was a change to the if-check, so the only place I could see a few tests would be `OffsetDateTimeSerTest`.

cowtowncoder: +1 for similar teest for `OffsetTimeSerTest`. Was wondering about whether config really should be through `SerializationFeature` as it only applies to this module (and since I hope to address "DateTimeConfig" for 2.14).
But given timing I guess let's go with `SerializationFeature` even if not optimal.

@vmeunier if you have a chance to add this test, I'm happy to merge the PR and can proceed with 2.13.0-rc1 very soon.

vmeunier: @kupci @cowtowncoder 
Just so you know, I can't add a test with a custom formatter for OffsetDateTime since every constructor is protected.

You said you didn't want to add a public one (such as the one in ZonedDateTimeSerializer). So i'm not changing anything retarding this, but I still added some tests (in both ZonedDateTimeSerTest and OffsetDateTimeSerTest).

vmeunier: Should I squash my commits ?
Do you have the autosquash ON on PR ?

cowtowncoder: I can squash them, no worries.

As to custom formatter, I think custom  _format_ (via `@JsonFormat` and/or config overrides) is good to test -- but no need to manually construct instances with formatters. So only things users do via public API (including annotations), unless there is specific case reported that we think is good to cover.

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
```diff
@@ -270,6 +273,64 @@ public void testSerializationAsStringWithZoneIdOn() throws Exception {
         assertEquals("The value is incorrect.", "\"" + DateTimeFormatter.ISO_ZONED_DATE_TIME.format(date) + "\"", value);
     }
 
+    @Test
+    public void testSerializationAsStringWithDefaultTimeZoneAndContextTimeZoneOnAndACustomFormatter() throws Exception {
+        ZonedDateTime date = ZonedDateTime.now(Z3);
+        // With a custom DateTimeFormatter without a ZoneId.
+        String value = MAPPER.registerModule(new SimpleModule().addSerializer(new ZonedDateTimeSerializer(FORMATTER_WITHOUT_ZONEID))).writer()
```
cowtowncoder, line 280: As I mentioned in a comment, instead of constructing formatter, it'd make more sense to use `@JsonFormat` annotation or config overrides -- that's the expected usage. But this test is fine for now.

cowtowncoder:
Ok looks good; use of formatter is fine for now as per my comment. So this is good to merge wrt changes.

But one thing I should have asked earlier but I think I missed earlier: CLA. If I haven't yet asked and gotten one (I don't think I have but just in case), it's this 1-page doc:

and is usually easiest to print, fill & sign, scan/photo, email to `info` at fasterxml dot com. Once I receive it I can proceed with merging. This is only needed once and works for all future contributions.

Looking forward to merging this improvement!

vmeunier:
@cowtowncoder 
I have sent you the signed PDF

Issue #226, opened 2021-08-20
Broken links in README


skagedal:
From the page https://github.com/FasterXML/jackson-modules-java8/tree/master/datetime there are some broken links to javadoc pages. 

For example to http://fasterxml.github.com/jackson-databind/javadoc/2.2.0/com/fasterxml/jackson/databind/SerializationFeature.html#WRITE_DATES_AS_TIMESTAMPS

PR #227, opened 2021-08-20
Fix broken links to Serialization / Deserialization

kupci:
Fix for #226. This fixes the broken links using Github's automatic anchors, however it only links to the header, should it be more granular?


cowtowncoder:
I think this is a great start -- let's start with it.

Issue #234, opened 2022-04-12
Deserialize enum with @JsonCreator failed: argument type mismatch


Brozen:
version: 2.10.4
JDK: AdoptOpenJDK (HotSpot) 11.0.9
OS: macOS Monterey
CPU: 2GHz 4C Interl Core i5

I'm serializing and deserializing enums with `@JsonValu` and `@JsonCreator`. The code works in a long term, but after I register `ParameterNamesModule`, I got a exception when deserializing（serializing is OK）, stack trace likes:

```log
com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `ParameterNameIgnoreTest$MeetingStatus`, problem: argument type mismatch
 at [Source: (String)"{"id":1000,"status":2}"; line: 1, column: 21] (through reference chain: ParameterNameIgnoreTest$Meeting["status"])
	at com.fasterxml.jackson.databind.exc.ValueInstantiationException.from(ValueInstantiationException.java:47)
	at com.fasterxml.jackson.databind.DeserializationContext.instantiationException(DeserializationContext.java:1735)
	at com.fasterxml.jackson.databind.DeserializationContext.handleInstantiationProblem(DeserializationContext.java:1109)
	at com.fasterxml.jackson.databind.deser.std.FactoryBasedEnumDeserializer.deserialize(FactoryBasedEnumDeserializer.java:146)
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:288)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:151)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4218)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3214)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3182)
	at ParameterNameIgnoreTest.test(ParameterNameIgnoreTest.java:113)
         ...... 
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #235, opened 2022-04-12
optimize help ignore parameter name parse

Brozen:
add an annotation `@JsonParamNameIgnore` to help ignore parameter name parse. And help resolve issue [#234](https://github.com/FasterXML/jackson-modules-java8/issues/234)


cowtowncoder:
I'd first like to make sure that my suggestion in #234 (addition of `mode = Mode.DELEGATING` for `@JsonCreator`) does not work -- that would be my primary suggestion here.
I don't really like the idea of having to use annotations for ignoring JDK constructor parameter names, although I can see why that would allow working around the problem.

cowtowncoder:
Closing as per suggested workaround.

Issue #236, opened 2022-04-12
Make use of BigDecimalParser (jackson-core) via `NumberInput`


pjfanning:
* also use BigDecimal.valueOf(long) where possible (cf https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#valueOf(long))

cowtowncoder:
Included in 2.14.0.

PR #237, opened 2022-04-12
use BigDecimalParser

pjfanning:
https://github.com/FasterXML/jackson-modules-java8/issues/236

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/DecimalUtils.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/DecimalUtils.java
@@ -16,6 +16,8 @@
 
 package com.fasterxml.jackson.datatype.jsr310;
 
+import com.fasterxml.jackson.core.io.BigDecimalParser;
+
 import java.math.BigDecimal;
 import java.time.Instant;
 import java.util.function.BiFunction;
@@ -92,7 +94,7 @@ public final class DecimalUtils
             }
             return BigDecimal.valueOf(seconds).setScale(9);
         }
-        return new BigDecimal(toDecimal(seconds, nanoseconds));
+        return BigDecimalParser.parse(toDecimal(seconds, nanoseconds));
     }
 
     /**
@@ -104,7 +106,7 @@ public final class DecimalUtils
         // !!! 14-Mar-2016, tatu: Somewhat inefficient; should replace with functionally
         //   equivalent code that just subtracts integral part? (or, measure and show
         //   there's no difference and do nothing... )
-        return value.subtract(new BigDecimal(integer)).multiply(ONE_BILLION).intValue();
+        return value.subtract(BigDecimal.valueOf(integer)).multiply(ONE_BILLION).intValue();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pjfanning:
@cowtowncoder I got this error in the CI build (java 11 only)

```
Error:    ZonedDateTimeSerTest.testSerializationAsTimestamp01NegativeSecondsWithDefaults:117 The value is not correct. expected:<1969-04-13T05:05:38.599Z[Etc/UTC]> but was:<1969-04-13T05:05:38.599Z[UTC]>
```

Seems unrelated to my change

kupci (approved):

kupci:
It does seem unrelated, especially given Java 8 and Java 14 builds run successfully. Assuming it fails without your changes, and if so, need to figure out if this was some change in Java 11, that then got restored in Java 14..

GedMarc:
change the test suite from 14 to 17 and 18 please, the results off 11 are correct here - 

14 is not the correct version to use on this due to 

```
Accounting Currency Format Support

Currency format instances with accounting style, in which the amount is formatted in parentheses in some locales, can be obtained by calling NumberFormat.getCurrencyInstance(Locale)

```
and 

```
core-libs/java.text
Plural Support in CompactNumberFormat

java.text.CompactNumberFormat is now capable of dealing with plural forms. For example, the number 2,000,000 is formatted to "2 Millionen" in LONG style, whereas 1,000,000 to "1 Million" in the German language.

```
And more, as an interim release a few options were tried to solve base issues with numbers (specifically negative) and the "compacting" and ISO- formula usage for faster calculations ;)

GedMarc:
please ensure jdk 11 is post update 9 in the test case - 

https://www.oracle.com/java/technologies/javase/11-0-9-bugfixes.html

pjfanning:
will rebase this once https://github.com/FasterXML/jackson-modules-java8/pull/239 is merged - should tidy up the test issue

GedMarc:
@cowtowncoder check comments on #239 - regarding the failures

Are you happy with a trampoline to the big decimal parser?  
If so, I would like to centralize the java.time ser/der to OffsetTimeSerializer and trampoline through that as it is also my preferred mechanism, 
https://github.com/GedMarc/GuiceInjection/blob/master/src/main/java/com/guicedee/guicedinjection/json/LocalDateDeserializer.java#L46

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/DecimalUtils.java
```diff
@@ -92,7 +94,7 @@ public static BigDecimal toBigDecimal(long seconds, int nanoseconds)
             }
             return BigDecimal.valueOf(seconds).setScale(9);
         }
-        return new BigDecimal(toDecimal(seconds, nanoseconds));
+        return BigDecimalParser.parse(toDecimal(seconds, nanoseconds));
```
cowtowncoder, line 97: Ok, one suggestion here (which I probably should have made earlier) -- I think this should call

    NumberInput.parseBigDecimal(....)

instead of directly referencing `BigDecimalParser`.

cowtowncoder:
@GedMarc what do you mean wrt "trampoline" here? (I recall it being used to refer to double indirection, just not sure what that'd be here)

Otherwise changes look good although instead of direct ref to `BigDecimalParser`, should go via `NumberInput` wrapper (just to encapsulate details).

PR #239, opened 2022-04-12
run builds on java 17 and 18

pjfanning:
* use temurin jdks because they support java 18
* update mockito to support newer jdks
* adjust one ZonedDateTime test because the toString check was not reliable


.github/workflows/main.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: ['8', '11', '17', '18']
+        java_version: ['8', '11', '17']
```
GedMarc, line 25: addons:
  apt:
    packages:
      - openjdk-18-jdk

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
```diff
@@ -114,7 +114,14 @@ public void testSerializationAsTimestamp01NegativeSecondsWithDefaults() throws E
...
         String serialized = MAPPER.writeValueAsString(original);
         ZonedDateTime deserialized = MAPPER.readValue(serialized, ZonedDateTime.class);
-        assertEquals("The value is not correct.",  original, deserialized);
+        assertEquals("The day is not correct.", original.getDayOfMonth(), deserialized.getDayOfMonth());
+        assertEquals("The month is not correct.", original.getMonthValue(), deserialized.getMonthValue());
+        assertEquals("The year is not correct.", original.getYear(), deserialized.getYear());
+        assertEquals("The hour is not correct.", original.getHour(), deserialized.getHour());
+        assertEquals("The hour is not correct.", original.getMinute(), deserialized.getMinute());
+        assertEquals("The hour is not correct.", original.getSecond(), deserialized.getSecond());
+        assertEquals("The hour is not correct.", original.getNano(), deserialized.getNano());
```
GedMarc, line 123: nano

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
```diff
@@ -114,7 +114,14 @@ public void testSerializationAsTimestamp01NegativeSecondsWithDefaults() throws E
         ZonedDateTime original = ZonedDateTime.parse("Apr 13 1969 05:05:38.599 UTC", dtf);
         String serialized = MAPPER.writeValueAsString(original);
         ZonedDateTime deserialized = MAPPER.readValue(serialized, ZonedDateTime.class);
-        assertEquals("The value is not correct.",  original, deserialized);
+        assertEquals("The day is not correct.", original.getDayOfMonth(), deserialized.getDayOfMonth());
+        assertEquals("The month is not correct.", original.getMonthValue(), deserialized.getMonthValue());
+        assertEquals("The year is not correct.", original.getYear(), deserialized.getYear());
+        assertEquals("The hour is not correct.", original.getHour(), deserialized.getHour());
+        assertEquals("The hour is not correct.", original.getMinute(), deserialized.getMinute());
+        assertEquals("The hour is not correct.", original.getSecond(), deserialized.getSecond());
```
GedMarc, line 122: second

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
```diff
@@ -114,7 +114,14 @@ public void testSerializationAsTimestamp01NegativeSecondsWithDefaults() throws E
         ZonedDateTime original = ZonedDateTime.parse("Apr 13 1969 05:05:38.599 UTC", dtf);
         String serialized = MAPPER.writeValueAsString(original);
         ZonedDateTime deserialized = MAPPER.readValue(serialized, ZonedDateTime.class);
-        assertEquals("The value is not correct.",  original, deserialized);
+        assertEquals("The day is not correct.", original.getDayOfMonth(), deserialized.getDayOfMonth());
+        assertEquals("The month is not correct.", original.getMonthValue(), deserialized.getMonthValue());
+        assertEquals("The year is not correct.", original.getYear(), deserialized.getYear());
+        assertEquals("The hour is not correct.", original.getHour(), deserialized.getHour());
+        assertEquals("The hour is not correct.", original.getMinute(), deserialized.getMinute());
```
GedMarc, line 121: minute

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
```diff
@@ -114,7 +114,14 @@ public void testSerializationAsTimestamp01NegativeSecondsWithDefaults() throws E
         ZonedDateTime original = ZonedDateTime.parse("Apr 13 1969 05:05:38.599 UTC", dtf);
```
GedMarc, line 114: jsr310 is deprecated for jackson json - @pjfanning  

@cowtowncoder are there still updates to support this?

GedMarc (approved):

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
GedMarc: Oops

.github/workflows/main.yml
pjfanning: where does this go - github actions yaml is really badly documented - I can't find any examples like this to copy from

GedMarc:
Okie my bad - took a while to find sorry 

https://bugs.openjdk.java.net/browse/JDK-8196398

"wont-fix"

reason is because
```
"Etc/UTC" is categorized as "region-based ID" according to the ZoneId class description:

"The third type of ID are region-based IDs. A region-based ID must be of two or more characters, and not start with 'UTC', 'GMT', 'UT' '+' or '-'. Region-based IDs are defined by configuration, see ZoneRulesProvider. The configuration focuses on providing the lookup from the ID to the underlying ZoneRules."

This is different from "UTC". Even though TZ database regards UTC and Etc/UTC the same, (although very hypothetically) a ZoneRulesProvider can provide its own rules for Etc/UTC.
```

, use normalize()  (and each value cause that is great) as well as the string rep  ```Possible workaround for this is to compare ZoneIds with normalized() method.```

.github/workflows/main.yml
GedMarc: oh shouldn't use it as a java version direclty, as it is an apt install on the o.s. if applicable :) I think you're right, perhaps better to wait for travis to officially have an image available, testing it now it also needs the apt set-18-default apt as well,

GedMarc (approved):

GedMarc:
@cowtowncoder there's a funny here, only reason JDK 8 passes, is because it is before update 152 on travis image / everyone using a later version of jdk8 would also experience the difference (tested)

The Temurin switch I do agree with, to make sure it runs with the latest builds, but as a test case, this could become flaky on backwards compatibility to jdk8 <= 151

Ref : https://bugs.openjdk.java.net/browse/JDK-8196398 
The fact it was passing on JDK 8 is actually the bug

parameter-names/src/test/java/com/fasterxml/jackson/module/paramnames/ParameterNamesAnnotationIntrospectorTest.java
```diff
@@ -9,16 +9,16 @@
 import org.junit.runner.RunWith;
 import org.mockito.BDDMockito;
 import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnitRunner;
```
GedMarc, line 12: is this needed on any of the other tests with the mockito version bump?

pjfanning: this appears to be the only test in this code set that uses mockito

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/ser/ZonedDateTimeSerTest.java
cowtowncoder: Yeah I think that despite deprecation, 2.x should make reasonable effort to keep it working still.

parameter-names/src/test/java/com/fasterxml/jackson/module/paramnames/ParameterNamesAnnotationIntrospectorTest.java
cowtowncoder: Would be great to remove the dependency if it's the only usage. But not sure how much work that would be.

cowtowncoder:
@GedMarc @pjfanning is this good to go? Happy to merge if so.

parameter-names/src/test/java/com/fasterxml/jackson/module/paramnames/ParameterNamesAnnotationIntrospectorTest.java
pjfanning: the test seems to fail with that this

pjfanning: the older (original) mockito jar did not support latest java versions

pjfanning:
I'm happy with this

PR #240, opened 2022-04-13
LocalDateDeserializer should reflect coercionConfig

maciekdeb:
Currently LocalDateDeserializer does not take into account the following configuration:
`setCoercion(CoercionInputShape.Integer, CoercionAction.Fail);`

### DIFF
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
@@ -1,10 +1,5 @@
 package com.fasterxml.jackson.datatype.jsr310.deser;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
-
 import java.io.IOException;
 import java.time.Instant;
 import java.time.LocalDate;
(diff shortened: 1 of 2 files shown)


datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
```diff
@@ -476,7 +478,7 @@ public void testLenientDeserializeFromNumberInt() throws Exception {
         mapper.configOverride(LocalDate.class)
                         .setFormat(JsonFormat.Value.forShape(JsonFormat.Shape.NUMBER_INT));
 
-        assertEquals("The value is not correct.", LocalDate.of(1970, Month.MAY, 04),
+        assertEquals("The value is not correct.", LocalDate.of(1970, Month.MAY, 4),
```
kupci, line 481: Odd, I wonder why it used to be 04 vs. 4.. anyway, this cleanup looks fine to me.

kupci (approved):
Looks good to me, we can look for @cowtowncoder for a final review. Thanks for the PR!

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
cowtowncoder: Maybe seemed more readable by original author? (I didn't write it)

cowtowncoder:
Looks good to me. Thank @maciekdeb !

Only one more thing I need before merging: if you have not sent a CLA before (I saw one with similar name but not bound to this github handle?), I'd need one from here:

and the usual way is to print it, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I have CLA I can merge this for inclusion in 2.14.0!

Thank you once again for the contribution!

maciekdeb:
Thanks @kupci & @cowtowncoder.

I've just sent the CLA to this email address. 

I guess other date classes also misses it, when I have more time I can review them, I changed only LocalDate for now as this was the one I stumbled upon.

cowtowncoder:
@maciekdeb Thank you again! Yes, I think all others could and should be retrofitted too -- that would be valuable contribution. Same also goes for [`jackson-datatype-joda`](https://github.com/FasterXML/jackson-datatype-joda/) as well.

PR #241, opened 2022-04-27
Fix InstantSerializer ignoring the JsonFormat shape

KaseiFR:


cowtowncoder:
First of all: thank you for providing this patch! It makes sense.

Second: would it be possible to do PR against `2.14` instead of `master`? Master is for upcoming Jackson 3.0, and getting this released would occur much sooner via 2.14 release.
I can handle merging from 2.14 to master but the reverse is bit more difficult due to various reasons.

KaseiFR:
Here's the new PR: #242
I knew something was fishy when I couldn't find the config methods on `ObjectMapper` during testing: it's finally immutable !

PR #246, opened 2022-07-26
minor tweak to pass the JsonParser instance 

pjfanning:
Passing JsonParser instance allows features to be checked - the fast double parser feature being one that could be set


cowtowncoder:
Good catch, thanks!

PR #255, opened 2022-12-05
Change InstantSerializerBase to format dates in the same way as DateTimeSerializerBase

rotilho:
This fix an issue I have with jsonSchema where the schema generated for Dates differ from the Java Time ones.

I believe this also may affect other applications that relies on jackson formats.


cowtowncoder:
Ok, first of all: thank you for contribution!

Second, a couple of requests:

1. Please add actual description (seems like this would be for Swagger generation or something)?
2. Any chance of a unit test? (I don't remember if schema generation aspects have existing tests -- if none then that's ok but ideally changes would be checked by a test)

rotilho:
@cowtowncoder thanks for reviewing it, I just added a description.

Regarding the tests I couldn't find any similar tests or way where schema can be validate. It's my first time debugging the Jackson internals so I may just missing something obvious.

If you have an example somewhere I will be happy to add the unit tests.

cowtowncoder:
Ok thank you for adding description!

One thing I was wondering was whether there might be backwards-compatibility concerns, but it looks like you are only adding a new property, not changing. So maybe that is fine. But definitely needs to go in 2.15 not 2.14.x patch.

As to tests, there is existing test class:

    datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/misc/DateTimeSchemasTest.java

which I hope helps. It doesn't have to be anything big, just verify the fix in a minimal way (or not minimal if you prefer :) ).

cowtowncoder (approved):

cowtowncoder (approved):

PR #257, opened 2022-12-20
Correct link to wiki

reftel:
As rendered on GitHub (where the wiki link can link to something), there is a "blob" directory in the path that needs to be skipped in order for the wiki link to work.


cowtowncoder (approved):

cowtowncoder:
Thanks!

PR #260, opened 2023-01-03
JavaTimeModule is default in JPMS module-info. Fixes #259

### LINKED ISSUE #259: Wrong module auto-registered when using JPMS
As I understand module `com.fasterxml.jackson.datatype.jsr310.JSR310Module` is left for compatibility and `com.fasterxml.jackson.datatype.jsr310.JavaTimeModule` should be default one. However when using `ObjectMapper.findAndRegisterModules()` with JPMS in Java9+ `JSR310Module` is loaded as declared in module-info.java.

It should be changed as in #126 to maintain same behavior between modularized and non-modularized environments.

karbi:


cowtowncoder:
@karbi Makes sense, but for compatibility reasons I think this has to go in a minor version and not a patch. So could you please re-base against `2.15` branch?

cowtowncoder:
Created #262 to target 2.15 -- will merge that. Thank you @karbi !

PR #266, opened 2023-02-08
Optimize InstantDeserializer replaceZeroOffsetAsZIfNecessary

schlosna:
While reviewing some JFRs from some services that heavily use Jackson for deserializing timestamps, I noticed that `com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.replaceZeroOffsetAsZIfNecessary(String)` was allocating a lot of `int[]` from its use of `java.util.regex.Pattern.matcher(CharSequence)` to determine whether an input String is a zero offset and if so replace it with `Z`.

image

I would like to propose an alternative implementation to avoid expensive regex matcher allocations when replacing zero offset with 'Z'.

Using a [JMH benchmark with a variety of generated timestamps](https://gist.github.com/schlosna/abe63391e890a672a88330db13989db8) I see the following results on x64 (~7x speedup) and aarch64 (~2.5x speedup):

#### OpenJDK 17.0.6 on x64 Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
```
Benchmark                           Mode  Cnt    Score   Error  Units
InstantDeserializerBenchmark.index  avgt    5   19.543 ± 1.384  ns/op
InstantDeserializerBenchmark.regex  avgt    5  155.081 ± 3.234  ns/op
```

#### OpenJDK 17.0.6 on aarch64 Apple M1 Pro
```
Benchmark                           Mode  Cnt   Score   Error  Units
InstantDeserializerBenchmark.index  avgt    5  16.855 ± 0.537  ns/op
InstantDeserializerBenchmark.regex  avgt    5  58.155 ± 1.444  ns/op
```

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -55,13 +55,6 @@ public class InstantDeserializer<T extends Temporal>
     private static final long serialVersionUID = 1L;
 
     /**
-     * Constants used to check if the time offset is zero. See [jackson-modules-java8#18]
-     *
-     * @since 2.9.0
-     */
-    private static final Pattern ISO8601_UTC_ZERO_OFFSET_SUFFIX_REGEX = Pattern.compile("\\+00:?(00)?$");
-
-    /**
      * Constants used to check if ISO 8601 time string is colonless. See [jackson-modules-java8#131]
      *
      * @since 2.13
@@ -226,7 +219,7 @@ public class InstantDeserializer<T extends Temporal>
                 return (T) parser.getEmbeddedObject();
 
             case JsonTokenId.ID_START_ARRAY:
-            	return _deserializeFromArray(parser, context);
+                return _deserializeFromArray(parser, context);
         }
         return _handleUnexpectedToken(context, parser, JsonToken.VALUE_STRING,
                 JsonToken.VALUE_NUMBER_INT, JsonToken.VALUE_NUMBER_FLOAT);
@@ -337,9 +330,28 @@ public class InstantDeserializer<T extends Temporal>
     private String replaceZeroOffsetAsZIfNecessary(String text)
     {
         if (replaceZeroOffsetAsZ) {
-            return ISO8601_UTC_ZERO_OFFSET_SUFFIX_REGEX.matcher(text).replaceFirst("Z");
+            return replaceZeroOffsetAsZ(text);
         }
 
+        return text;
+    }
+
+    private static String replaceZeroOffsetAsZ(String text)
+    {
+        int plusIndex = text.lastIndexOf('+');
+        if (plusIndex < 0) {
+            return text;
+        }
+        int maybeOffsetIndex = plusIndex + 1;
+        int remaining = text.length() - maybeOffsetIndex;
+        if (remaining < 2 || remaining == 3 || remaining > 5) {
+            return text;
+        }
+
+        String maybeOffset = text.substring(maybeOffsetIndex);
+        if ("00".equals(maybeOffset) || "0000".equals(maybeOffset) || "00:00".equals(maybeOffset)) {
+            return text.substring(0, plusIndex) + 'Z';
+        }
         return text;
     }
 
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -344,13 +344,19 @@ public class InstantDeserializer<T extends Temporal>
         }
         int maybeOffsetIndex = plusIndex + 1;
         int remaining = text.length() - maybeOffsetIndex;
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):

cowtowncoder:
Excellent, thank you @schlosna! I especially appreciate jmh benchmark here.

Just one small thing before I can merge this for inclusion in 2.15.0: unless I have asked for a CLA, I'd need it from:

and the usual way is to print it, fill & sign, scan/photo, email to `info` at fasterxml dot com.
This only needs to be done once before the first contribution so if you already sent it once let me know.

Thank you again & looking forward to merging this PR!

cowtowncoder:
Ok, one oddity: new test cases fail on JDK 8 and 11 (see CI run; I had to enable it since it's your first contribution). Possibly related to handling (or not) of colon in time offset parts?

schlosna:
Thanks for the quick review @cowtowncoder ! I'll get the CLA reviewed & signed hopefully tomorrow.

I removed the commit with additional test cases as I forgot [those will fail on JDK 8 & 11](https://github.com/FasterXML/jackson-modules-java8/actions/runs/4125352620/jobs/7137027552) (the test cases also fail on JDK 8 & 11 applied to `origin/2.15`)

schlosna:
After some needed sleep, I realized why those tests pass on JDK 17+ but did not pass on JDK 8 & 11 and it is the reason we're replacing the zero offset with `Z` in the first place -- before JDK 12 `DateTimeFormatter.ISO_INSTANT` didn't handle offsets. This was added by [JDK-8166138](https://bugs.openjdk.org/browse/JDK-8166138) & https://github.com/openjdk/jdk/commit/f5d19b9109c7a0a40a85ca022d3089230ed1456b

I've updated the test cases in `InstantDeserTest` to assume JDK 12+ for those cases, and validated that the `OffsetDateTimeDeserTest` and `ZonedDateTimeDeserTest` also cover `replaceZeroOffsetAsZIfNecessary`.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -337,12 +330,31 @@ private ZoneId getZone(DeserializationContext context)
...
+            return text;
+        }
+        int maybeOffsetIndex = plusIndex + 1;
+        int remaining = text.length() - maybeOffsetIndex;
+        if (remaining < 2 || remaining == 3 || remaining > 5) {
+            return text;
+        }
+
+        String maybeOffset = text.substring(maybeOffsetIndex);
+        if ("00".equals(maybeOffset) || "0000".equals(maybeOffset) || "00:00".equals(maybeOffset)) {
```
carterkozak, line 352: Can we optimize this using `regionMatches`?
perhaps using something along the lines of
```java
switch (remaining) {
    case 2:
       return text.regionmatches..."00" ? text.substring(0, plusIndex) + 'Z' : text;
     case 4:
        ..etc
     case 5:
       ..etc
 }
 return text;
```

schlosna: Seems reasonable to avoid the substring allocation, updated to use `regionMatches` & `switch` and reran JMH and tested locally with JDKs 8, 11, 17.

```
Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
JDK 17.0.5, OpenJDK 64-Bit Server VM, 17.0.5+8-LTS
Benchmark                           Mode  Cnt    Score   Error  Units
InstantDeserializerBenchmark.index  avgt    5   20.003 ± 0.486  ns/op
InstantDeserializerBenchmark.regex  avgt    5  156.174 ± 7.654  ns/op
```

```
Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
JDK 11.0.16.1, OpenJDK 64-Bit Server VM, 11.0.16.1+9-LTS
Benchmark                           Mode  Cnt    Score   Error  Units
InstantDeserializerBenchmark.index  avgt    5   20.818 ± 1.267  ns/op
InstantDeserializerBenchmark.regex  avgt    5  192.248 ± 7.341  ns/op
```

```
Apple M1 Pro aarch64
JDK 17.0.6, OpenJDK 64-Bit Server VM, 17.0.6+10-LTS
Benchmark                           Mode  Cnt   Score   Error  Units
InstantDeserializerBenchmark.index  avgt    5  17.498 ± 0.677  ns/op
InstantDeserializerBenchmark.regex  avgt    5  58.096 ± 1.330  ns/op
```

```
Apple M1 Pro aarch64
Benchmark                           Mode  Cnt   Score   Error  Units
InstantDeserializerBenchmark.index  avgt    5  16.176 ± 7.842  ns/op
InstantDeserializerBenchmark.regex  avgt    5  63.201 ± 5.668  ns/op
```

cowtowncoder:
Ok sounds good; will merge as soon as we get CLA done. Good work everyone!

schlosna:
I'll submit CLA ASAP, waiting on review right now

schlosna:
CLA submitted via email

cowtowncoder (approved):

cowtowncoder:
Thank you @schlosna -- merged to be included in 2.15.0!

Issue #263, opened 2023-01-09
Add `JavaTimeFeature.ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS` to allow parsing quoted numbers when using a custom DateTimeFormatter


mpkorstanje:
I have a scenario where I would like to de-serialize epoch milis, and serialize with a `@JsonFormat` annotation. In essence:

```java
  final ObjectMapper mapper = new ObjectMapper()
      .disable(DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS)
      .disable(SerializationFeature.WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS)
      .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
      .setTimeZone(TimeZone.getTimeZone("UTC"))
      .findAndRegisterModules();
  final Instant instant = Instant.ofEpochMilli(123456);

  @Test
  public void test() throws JsonProcessingException {
    assertEquals("{\"time\":\"1970-01-01T00:02:03.456+0000\"}", mapper.writeValueAsString(new TimeHolderWithFormat(instant)));
    assertEquals(instant, mapper.readValue("{\"time\":123456}", TimeHolder.class).time);
    assertEquals(instant, mapper.readValue("{\"time\":\"123456\"}", TimeHolder.class).time);
    assertEquals(instant, mapper.readValue("{\"time\":123456}", TimeHolderWithFormat.class).time);
    assertEquals(instant, mapper.readValue("{\"time\":\"123456\"}", TimeHolderWithFormat.class).time);
  }

  
  static class TimeHolder {
    private Instant time;
    public void setTime(Instant time) {
      this.time = time;
    }
    // Getter ommited
  }

  static class TimeHolderWithFormat {
    private Instant time;
    // Constructors ommited
    @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSZ")
    public void setTime(Instant time) {
      this.time = time;
    }
    // Getter ommited
  }
```

This will fail on the last assertion with:

```
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.Instant` from String "123456": Failed to deserialize java.time.Instant: (java.time.format.DateTimeParseException) Text '123456' could not be parsed at index 0
[...]

PR #269, opened 2023-03-15
Add feature toggle to read numeric strings as numeric timestamps

mpkorstanje:
There is an apparent inconsistency in the way Jackson de-serializes numbers that are shaped as a string into an instant.

 * When not providing a custom date format, quoted numbers are treated as epoch milis/nanos
 * When not providing a custom date, quoted numbers are assumed to be handled by the pattern.

There is however no way to construct a pattern that handles both ISO dates and epoch milis/nanos. This feature toggle allow numeric strings to be read as numeric timestamps.

Fixes: https://github.com/FasterXML/jackson-modules-java8/issues/263.

This PR is paired with https://github.com/FasterXML/jackson-databind/pull/3830.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -20,14 +20,16 @@ public enum JavaTimeFeature implements JacksonFeature
     NORMALIZE_DESERIALIZED_ZONE_ID(true),
 
     /**
-     * Feature that controls whether numeric strings may be interpreted as numeric
-     * timestamps (enabled) or not (disabled), in addition to an explicitly
-     * defined pattern.
+     * Feature that controls whether stringified numbers (Strings that without
+     * quotes would be legal JSON Numbers) may be interpreted as
+     * timestamps (enabled) or not (disabled), in case where there is an
+     * explicitly defined pattern ({@code DateTimeFormatter}) for value.
      * <p>
-     * Note that when a pattern is not explicitly defined numeric strings are
-     * interpreted as a numeric timestamp.
+     * Note that when the default pattern is used (no custom pattern defined),
+     * stringified numbers are always accepted as timestamps regardless of
+     * this feature.
      */
-    ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS(false)
+    ALWAYS_ALLOW_STRINGIFIED_DATE_TIMESTAMPS(false)
     ;
 
   /**
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -20,14 +20,14 @@ public enum JavaTimeFeature implements JacksonFeature
     NORMALIZE_DESERIALIZED_ZONE_ID(true),
 
     /**
-     * Feature that controls whether numeric strings are interpreted as numeric
-     * timestamps (enabled) nor not (disabled) in addition to an explicitly
+     * Feature that controls whether numeric strings may be interpreted as numeric
+     * timestamps (enabled) or not (disabled), in addition to an explicitly
      * defined pattern.
      * <p>
      * Note that when a pattern is not explicitly defined numeric strings are
      * interpreted as a numeric timestamp.
      */
-    READ_NUMERIC_STRINGS_AS_DATE_TIMESTAMP(false);
+    ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS(false);
 
     /**
      * Whether feature is enabled or disabled by default.
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -17,7 +17,17 @@ public enum JavaTimeFeature implements JacksonFeature
      * Default setting is enabled, for backwards-compatibility with
      * Jackson 2.15.
      */
-    NORMALIZE_DESERIALIZED_ZONE_ID(true);
+    NORMALIZE_DESERIALIZED_ZONE_ID(true),
+
+    /**
+     * Feature that controls whether numeric strings are interpreted as numeric
+     * timestamps (enabled) nor not (disabled) in addition to an explicitly
+     * defined pattern.
+     * <p>
+     * Note that when a pattern is not explicitly defined numeric strings are
+     * interpreted as a numeric timestamp.
+     */
+    READ_NUMERIC_STRINGS_AS_DATE_TIMESTAMP(false);
 
     /**
      * Whether feature is enabled or disabled by default.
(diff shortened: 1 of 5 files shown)


mpkorstanje:
> [Contributing - Paper Work](https://github.com/FasterXML/jackson/blob/master/CONTRIBUTING.md)
> All you need to do is to download the CLA document, print it, fill and sign, scan (or take photo on your phone) and email that copy to info at fasterxml dot com.

One moment while I print liquid gold on dead tree.

cowtowncoder:
Ok: as per my comments, I'd like to use a module configuration setting here.

But looking at the way deserializers are implemented, this appears a bit tricky: not so much wrt configuring `JavaTimeModule`, but in passing configuration setting(s) from module to `InstantDeserializer`... some refactoring would be needed, so that instead of pre-created instances, we'd create instances when registering deserializers.
Alas, `InstantDeserializer.INSTANT` and 2 others are `public static` singletons so we must leave something like that.
Let me think about this a bit.

cowtowncoder:
Ok, I found time to implement #281, and for that introduced `JavaTimeFeature` enumeration -- this now allows adding a setting for this PR as well. The main/only concern for me is just naming the feature to reflect semantics.

mpkorstanje:
Cool. I'll rework the PR to use that.

mpkorstanje:
> The main/only concern for me is just naming the feature to reflect semantics.

What do you think are the important semantics?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
```diff
@@ -17,7 +17,17 @@ public enum JavaTimeFeature implements JacksonFeature
...
+
+    /**
+     * Feature that controls whether numeric strings are interpreted as numeric
+     * timestamps (enabled) nor not (disabled) in addition to an explicitly
+     * defined pattern.
+     * <p>
+     * Note that when a pattern is not explicitly defined numeric strings are
+     * interpreted as a numeric timestamp.
+     */
+    READ_NUMERIC_STRINGS_AS_DATE_TIMESTAMP(false);
```
mpkorstanje, line 30: When it comes to `ALWAYS_ READ_NUMERIC_STRINGS_AS_DATE_TIMESTAMP` probably works.

The `ALWAYS` implies that even when disabled it happens some times.

There still is the implication that _any_ numeric string is read as a timestamp.

So then we get `ALWAYS_READ_DATE_TIMESTAMPS_FROM_NUMERIC_STRING`?

But your choice really. I'm not terribly attached to any of these.

cowtowncoder: :-)

Naming is hard!

So maybe something like

     ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS

would be good, to indicate that timestamps-as-strings are allowed, without claiming all JSON Strings must be timestamps (this will still only augment possibilities).

Let's try that one, get everything else straightened out, and this can still make it in 2.16.0.

cowtowncoder:
Looks like I got CLA too, great!

cowtowncoder:
@mpkorstanje I renamed the feature, made some other minor tweaks and so I think this is ready for merge from my POV.
But I wanted to let you have a look if you want some other changes (even better feature name? :) )
Just LMK either way, looking forward to getting this nice feature merged.

cowtowncoder (approved):

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
```diff
@@ -17,10 +17,20 @@ public enum JavaTimeFeature implements JacksonFeature
...
 
     /**
+     * Feature that controls whether numeric strings may be interpreted as numeric
+     * timestamps (enabled) or not (disabled), in addition to an explicitly
+     * defined pattern.
+     * <p>
+     * Note that when a pattern is not explicitly defined numeric strings are
+     * interpreted as a numeric timestamp.
+     */
+    ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS(false)
```
mpkorstanje, line 30: To be consistent with `READ_DATE_TIMESTAMPS_AS_NANOSECONDS` I reckon `ALWAYS_READ_STRINGIFIED_DATE_TIMESTAMPS` would be more accurate.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -130,17 +134,31 @@ public class InstantDeserializer<T extends Temporal>
...
+     *
+     * @since 2.16
      */
     protected final boolean _normalizeZoneId;
 
+    /**
+     * Flag set from
+     * {@link com.fasterxml.jackson.datatype.jsr310.JavaTimeFeature#ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS}
+     * to determine whether numeric strings are interpreted as numeric timestamps
+     * (enabled) nor not (disabled) in addition to an explicitly defined pattern.
```
mpkorstanje, line 146: Should we replace "pattern" with "custom DateTimeFormatter" as you added to the changelog?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -92,7 +95,8 @@ public class InstantDeserializer<T extends Temporal>
             a -> ZonedDateTime.ofInstant(Instant.ofEpochSecond(a.integer, a.fraction), a.zoneId),
             ZonedDateTime::withZoneSameInstant,
             false, // keep zero offset and Z separate since zones explicitly supported
-            DEFAULT_NORMALIZE_ZONE_ID 
+            DEFAULT_NORMALIZE_ZONE_ID,
+            DEFAULT_READ_NUMERIC_STRINGS_AS_DATE_TIMESTAMP
```
mpkorstanje, line 99: Should this match the feature toggle name?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -130,17 +134,31 @@ public class InstantDeserializer<T extends Temporal>
...
 
+    /**
+     * Flag set from
+     * {@link com.fasterxml.jackson.datatype.jsr310.JavaTimeFeature#ALWAYS_ALLOW_STRINGIFIED_TIMESTAMPS}
+     * to determine whether numeric strings are interpreted as numeric timestamps
+     * (enabled) nor not (disabled) in addition to an explicitly defined pattern.
+     *
+     * @since 2.16
+     */
+    protected final boolean _readNumericStringsAsTimestamp;
```
mpkorstanje, line 150: Should this match the feature toggle name?

mpkorstanje:
Sorry for the delay. Yesterday was a bit full. 

Naming aside, it looks good to me.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
cowtowncoder: I'll add `DATE_` in there, good suggestion.

The reason I think "ALLOW" is important is because this is used in addition to -- but not instead of -- custom pattern.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
cowtowncoder: Yes, good point it should.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
cowtowncoder: Yes.

cowtowncoder:
Thank you for all the work here & improvement suggestions @mpkorstanje !

PR #278, opened 2023-06-20
refactor jsr310 desers to be more consistent in withXxx methods

### LINKED ISSUE #276: JSR310DateTimeDeserializerBase#withDateFormate implementations without this as base class
I was analyzing the sources and have found some strange things.

 * LocalTimeDeserializer
 * YearDeserializer
 * LocalDateTimeDeserializer
 * MonthDayDeserializer
 * OffsetTimeDeserializer
 * YearMonthDeserializer

classes don't use this as base when ovverriding `JSR310DateTimeDeserializerBase#withDateFormate` and therefore we can loose leniency and shape. What is the real intention of such code or is it a bug? If this is inteded behaviour, e.g. because implementations don't use leniency and shape and only care about formatter, there is another strange thing why does `withShape` is ignored by returning `this` almost everywhere but ignoring 'withLeniency' leads to new deser made?

[LocalTimeDeserializer as an example](https://github.com/FasterXML/jackson-modules-java8/blob/818afc1fada88b63d47bed15f558bcd131c1985e/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateTimeDeserializer.java#L64-L75)

raman-babich:
Fixes #276

@cowtowncoder if it LGTY, plese merge, otherwise just simply close the issue.

Thanks in advance.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -209,9 +209,6 @@ public class InstantDeserializer<T extends Temporal>
         return new InstantDeserializer<>(this, _formatter, leniency);
     }
 
-    @Override
-    protected InstantDeserializer<T> withShape(JsonFormat.Shape shape) { return this; }
-
     @SuppressWarnings("unchecked")
     @Override // @since 2.12.1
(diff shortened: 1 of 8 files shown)


cowtowncoder (approved):
LGTM!

cowtowncoder:
@raman-babich Thank you once again: these kinds of clean ups are very valuable so definitely appreciate PR here. I value pro-active maintenance of codebase to clean up cruft that accumulates due to addition of new features etc.

PR #287, opened 2023-11-18
#286: Get the pre 2.16 constructor of InstantDeserializer back

### LINKED ISSUE #286: Breaking change in `InstantDeserializer API between 2.15 and 2.16
Version 2.16 seems to have introduced a breaking change in the API for `InstantDeserializer`.

Since 6 years ago…. And up until 2.15, the [InstantDeserializer constructor](https://github.com/FasterXML/jackson-modules-java8/blob/cc5b7f030be0b2ff9641b3f79e7be16e5b3078a6/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java#L115C1-L121C42) looked like this:

```java
    protected InstantDeserializer(Class<T> supportedType,
            DateTimeFormatter formatter,
            Function<TemporalAccessor, T> parsedToValue,
            Function<FromIntegerArguments, T> fromMilliseconds,
            Function<FromDecimalArguments, T> fromNanoseconds,
            BiFunction<T, ZoneId, T> adjust,
            boolean replaceZeroOffsetAsZ)
```

However, since 2.16 and onwards, [it was changed](https://github.com/FasterXML/jackson-modules-java8/blob/badd731b300c42def95fc35bfaa3dc436f8e8796/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java#L156C1-L165C6) to this:

```java
    protected InstantDeserializer(Class<T> supportedType,
            DateTimeFormatter formatter,
            Function<TemporalAccessor, T> parsedToValue,
            Function<FromIntegerArguments, T> fromMilliseconds,
            Function<FromDecimalArguments, T> fromNanoseconds,
            BiFunction<T, ZoneId, T> adjust,
            boolean replaceZeroOffsetAsZ,
            boolean normalizeZoneId,
            boolean readNumericStringsAsTimestamp
    )
```

Two new `boolean` parameters added.

This breaks our custom subclass, that no longer finds it `super` constructor. To make things worse for us, the subclass is shared between multiple projects, that use different versions of Jackson, so we can't just upgrade and fix it for the latest version...

Can we have the old version (with defaults for the missing parameters) back, please. 😀 

If you like, I can make a PR, I just need to figure out what the defaults should be.

haraldk:
Fixes #286 

I just added the old constructor back, delegating to the new one with defaults as suggested. You might have some input on the deprecation message. 

I didn't add any tests, let me know if that's needed (not really sure what to test in that case..?). 

The PR is against 2.17, not sure if it should be against 2.16?

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -176,6 +176,24 @@ public class InstantDeserializer<T extends Temporal>
         _alwaysAllowStringifiedDateTimestamps = readNumericStringsAsTimestamp;
     }
 
+    /**
+     * @deprecated Since 2.16, use {@link #InstantDeserializer(Class,DateTimeFormatter,
+     * Function,Function,Function,BiFunction,boolean,boolean,boolean)} instead.
+     */
+    @Deprecated()
+    protected InstantDeserializer(Class<T> supportedType,
+                                  DateTimeFormatter formatter,
+                                  Function<TemporalAccessor, T> parsedToValue,
+                                  Function<FromIntegerArguments, T> fromMilliseconds,
+                                  Function<FromDecimalArguments, T> fromNanoseconds,
+                                  BiFunction<T, ZoneId, T> adjust,
+                                  boolean replaceZeroOffsetAsZ
+
+    ) {
+        this(supportedType, formatter, parsedToValue, fromMilliseconds, fromNanoseconds, adjust, replaceZeroOffsetAsZ,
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Ok thank you! So,

1. No need for tests, addition is fine (feel bit ambivalent on this one, wrt testing... could be convinced about need for test I guess)
2. Yes, needs to be for `2.16` if and when this is to be included in `2.16.1`; `2.17` branch is for 2.17.0 which will take months to release. Unfortunately it may be necessary to create new PR since Github UI doesn't have a good way to re-base (AFAIK -- although would be happy to be proven wrong here!).

cowtowncoder (commented):
Looks good! But alas, needs to be against `2.16` branch

cowtowncoder:
Recreated as #290 just to avoid extra hassle here; will be merging once CI passes.

Thank you @haraldk for reporting this, submitting PR!

PR #290, opened 2023-11-21
Fix #286; add back InstantDeserializer constructor from 2.15, deprecated

### LINKED ISSUE #286: Breaking change in `InstantDeserializer API between 2.15 and 2.16

cowtowncoder:
Recreated #287 to target 2.16 branch.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -153,6 +153,9 @@ public class InstantDeserializer<T extends Temporal>
      */
     protected final boolean _alwaysAllowStringifiedDateTimestamps;
 
+    /**
+     * @since 2.16
+     */
     protected InstantDeserializer(Class<T> supportedType,
             DateTimeFormatter formatter,
             Function<TemporalAccessor, T> parsedToValue,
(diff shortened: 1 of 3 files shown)


haraldk:
Thanks @cowtowncoder! 👍🏻

PR #292, opened 2023-12-01
274: add MonthDeserializer and JavaTimeFeature option

etrandafir93:


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JSR310Module.java
```diff
@@ -18,6 +18,7 @@
 
 import java.time.*;
 
+import com.fasterxml.jackson.core.json.PackageVersion;
```
cowtowncoder, line 21: Please remove this, that is the wrong `PackageVersion`: the real one is generated on this package.
(generated with `./mvnw clean compile`)

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.java
```diff
@@ -18,6 +18,7 @@
 
 import java.time.*;
 
+import com.fasterxml.jackson.core.json.PackageVersion;
```
cowtowncoder, line 21: Remove this as well.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override
+    protected JSR310DateTimeDeserializerBase<Month> withLeniency(Boolean leniency) {
+        return this;
+    }
+
+    public MonthDeserializer withFeatures(JacksonFeatureSet<JavaTimeFeature> features) {
+        return new MonthDeserializer(this._formatter, features.isEnabled(JavaTimeFeature.ONE_BASED_MONTHS));
+    }
+
+    static class MonthParsingException extends JsonProcessingException {
```
cowtowncoder, line 98: No, let's not create new exceptions (see earlier comment on alternative)

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+                }
+            }
+            try {
+                int monthNo = Integer.parseInt(monthText);
+                if (_oneBaseMonths) {
+                    return Month.of(monthNo);
+                }
+                return Month.values()[monthNo];
+            } catch (NumberFormatException nfe) {
+                throw new MonthParsingException(new DateTimeParseException("Cannot parse java.time.Month", monthText, 0, nfe));
```
cowtowncoder, line 71: Should probably include some more information on legal values. But more importantly, instead of using a one-off exception type, should use existing one (`InvalidFormatException` most likely).
`DeserializationContext` has methods and in particular:

```
    public Object handleWeirdStringValue(Class<?> targetClass, String value,
            String msg, Object... msgArgs)
```

should be used here as it allows catching by `DeserializationProblemHandler`, but also constructing and throwing exception otherwise.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+        if(parser.currentToken() == JsonToken.VALUE_NUMBER_INT) {
+            int monthNo = parser.getIntValue();
+            if (_oneBaseMonths) {
+                return Month.of(monthNo);
```
cowtowncoder, line 77: Does this throw an exception for values outside range? If so, should probably check, throw explicitly as expected (could use `DeserializationContext.handleWeirdNumberValue(...)`)

And could use matching test for out-of-range case(s).

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+            return Month.values()[monthNo];
```
cowtowncoder, line 79: Same as above, but more important as we don't want to get `ArrayIndexOutOfBundsException`...

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        return null;
```
cowtowncoder, line 81: Ah, no. :)

As year, `YearDeserializer` has something to reuse:

```
        return _handleUnexpectedToken(context, parser, JsonToken.VALUE_STRING, JsonToken.VALUE_NUMBER_INT);
```

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserTest.java
```diff
@@ -0,0 +1,230 @@
...
+
+        try {
+            strictMapper.readerFor(MAP_TYPE_REF).readValue("{\"date\":\"\"}");
+            fail("Should not pass");
+        } catch (MismatchedInputException e) {
+            verifyException(e, "not allowed because 'strict' mode set for");
+        }
+    }
+
+    private void expectFailure(String aposJson) throws Throwable {
```
cowtowncoder, line 212: I know existing tests use this helper method, but it's.... pretty bad. I wish I had time to fix all of these from the original implementation that was contributed.

But basically verification should check that:

1. Exception class is EXACTLY what we expect and
2. Exception message matches completely, or relevant part does

so this helper method can be avoided and verification inlined in test methods.

cowtowncoder:
@etrandafir93 First of all, thank you for this contribution! I added a few suggestions, mostly to change things that could be done in better way.

One practical thing before merging (once we get code all cleaned up!) is CLA: unless we have received one before, we'd need it from here:

and the usual way is to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.
This only needs to be done once, and CLA is good for all future contributions.

Thank you once again; looking forward to merging this useful new feature!

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JSR310Module.java
cowtowncoder: Actually, I'll go ahead and remove it from pr.

cowtowncoder:
@etrandafir93 I figured it might be quicker for me to make suggested changes instead of explaining what and how should be changed.
I hope you don't mind pushing changes to PR.

But one thing that is bit worse off now, and that needs to be figure out is this: should "Stringified" numbers be accepted like regular JSON numbers or not? Other deserializers only allow them for special case of backends that have `StreamReadCapability.UNTYPED_SCALARS` enabled (currently meaning XML and CSV), but not for typed formats.
So I changed code in a way that makes this work: if you feel strongly this is wrong you can change that; either way, tests need to be aligned with whether these are allowed.

I did not add bounds checks for numbers; those should be done etc.

etrandafir93:
> @etrandafir93 I figured it might be quicker for me to make suggested changes instead of explaining what and how should be changed. I hope you don't mind pushing changes to PR.
> 
> But one thing that is bit worse off now, and that needs to be figure out is this: should "Stringified" numbers be accepted like regular JSON numbers or not? Other deserializers only allow them for special case of backends that have `StreamReadCapability.UNTYPED_SCALARS` enabled (currently meaning XML and CSV), but not for typed formats. So I changed code in a way that makes this work: if you feel strongly this is wrong you can change that; either way, tests need to be aligned with whether these are allowed.
> 
> I did not add bounds checks for numbers; those should be done etc.

@cowtowncoder - sure, feel free to commit on by branch. It's the first time I'm looking into the code and I'm not very familiar with it, but here's another idea that will allow us to skip re-implementing this parsing: 
If the feature is OFF (default) we can delegate to the generic enum deserializer, making sure we are 100% backaward compatible.  If the feature is ON we can still rely on the enum generic deserializer, but then just return the previous month (so if generic deserializer returns Month.FEBRUARY, we map it to Month.JANUARY. Like a very simple decorator pattern. What do you think?

cowtowncoder:
@etrandafir93 Ah! Conceptually I like the idea a lot -- and it should be quite compatible with existing handling, too (since `EnumDeserializer` is currently used).

The challenge, then, becomes that of implementing delegating deserializer. A good starting point would be `DelegatingDeserializer` from `jackson-databind`. An instance would need to be registered via different callback: `BeanDeserializerModifier` method `modifyEnumDeserializer()` that would create delegating instance for specific case of type `Month.class` (passing default deserializer to `DelegatingDeserializer`).

And then `DelegatingDeserializer` implementation would need to override 2 `deserialize()` method and `deserializeWithType()` (ideally), to have special handling for 1-based case.

etrandafir93:
hello @cowtowncoder, I have made some changes to start using _BeanDeserializerModifier_  and I believe i'm on the right track. However, I'm not sure I've fully understood your instructions. Can you take a look at the latest commit? 

Is there a way of registering the _BeanDeserializerModifier_ for a specific type better than what I did with the if  `type.getRawClass() != Month.class`? Let me know what you think about the solution and, if it is ok, I'll move on with formatting, documentation and other small refactorings.

Thanks for the support and quick replies!

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public MonthEnumDeserializerModifier() {
+        this(false);
+    }
+
+    public MonthEnumDeserializerModifier(boolean oneBaseMonths) {
+        _oneBaseMonths = oneBaseMonths;
+    }
+
+    public MonthEnumDeserializerModifier withFeatures(JacksonFeatureSet<JavaTimeFeature> features) {
```
cowtowncoder, line 28: Let's change this to be passed in constructor: it's an internal implementation class, no need for fluent-style, can be modified without worries about compatibility.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
```diff
@@ -0,0 +1,60 @@
...
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.DeserializationConfig;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeFeature;
+
+
+public class MonthEnumDeserializerModifier extends BeanDeserializerModifier {
```
cowtowncoder, line 17: Typically modules have at most one such modifier, not type specific. So maybe name as `JavaTimeDeserializerModifier` to allow future expansion.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
```diff
@@ -0,0 +1,60 @@
...
+    public MonthEnumDeserializerModifier withFeatures(JacksonFeatureSet<JavaTimeFeature> features) {
+        boolean enabled = features.isEnabled(JavaTimeFeature.ONE_BASED_MONTHS);
+        if (enabled == _oneBaseMonths) {
+            return this;
+        }
+        return new MonthEnumDeserializerModifier(enabled);
+    }
+
+    @Override
+    public JsonDeserializer<?> modifyEnumDeserializer(DeserializationConfig config, JavaType type, BeanDescription beanDesc, JsonDeserializer<?> deserializer) {
```
cowtowncoder, line 37: Maybe name `deserializer` as "default deserializer"?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
```diff
@@ -0,0 +1,60 @@
...
+        if(type.getRawClass() != Month.class) {
```
cowtowncoder, line 38: Yes, something like this. Could simplify bit by 

    if (!type.hasRawClass(Month.class)) { ...

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
```diff
@@ -0,0 +1,60 @@
...
+            return deserializer;
+        }
+        return new JsonDeserializer<Enum>() {
```
cowtowncoder, line 41: Ok, this is too simplistic: my idea was to extend `DelegatingDeserializer` passing the default deserializer and overriding its `deserialize()` method -- ONLY handling special case of 1-based month and number, otherwise calling `super.deserialize()` (`getDelegatee().deserialize(...)` .

etrandafir93: I have applied your suggestions and a few other small changes. Can you take a look when you'll have some time?
Also, I haven't tackled the part about serialization yet - will we follow the same approach?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
etrandafir93: done

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthEnumDeserializerModifier.java
etrandafir93: renamed

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.time.Month;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.deser.std.DelegatingDeserializer;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+
```
cowtowncoder, line 12: Add `@since 2.17` in Javadoc here too

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+                throw new InvalidFormatException(parser, "Month.JANUARY value not allowed for 1-based Month.", zeroBaseMonth, Month.class);
+            } else {
+                return zeroBaseMonth.minus(1);
+            }
+        }
+        return zeroBaseMonth;
+    }
+
+    private boolean _isNumberAsString(String text, JsonToken token) throws IOException {
+        String regex = "[\"']?\\d{1,2}[\"']?";
```
cowtowncoder, line 40: Ok, couple of things: `parser.getText()` will return contents, so no quotes around it, can leave those out.

And make sure to create pre-compiled `Pattern` as static member of class: otherwise RegEx state machine is re-create for every call... and that's pretty expensive (and more importantly easily avoidable :) ).

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/MonthDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public MonthDeserializer(JsonDeserializer<Enum> defaultDeserializer, boolean oneBaseMonths) {
+        super(defaultDeserializer);
+        _enableOneBaseMonths = oneBaseMonths;
+    }
+
+    @Override
+    public Object deserialize(JsonParser parser, DeserializationContext context) throws IOException {
+        Month zeroBaseMonth = (Month) getDelegatee().deserialize(parser, context);
+        if (!_enableOneBaseMonths || zeroBaseMonth == null) {
```
cowtowncoder, line 24: I think I would actually change this so that if `! _enableOneBaseMonths`, or current token not `JsonToken.VALUE_NUMBER_INT` (and I guess "Stringified" number),  just call `delegate.deserialize()` and return value you get.

This is based on understanding that, I think, String name like `"FEBRUARY"` should NOT be changed to `JANUARY`? At least to me 0- vs 1-based logic only applies to numeric (and stringified numbers) case, not to textual.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/MonthSerializer.java
```diff
@@ -0,0 +1,36 @@
...
+        _enableOneBaseMonths = oneBaseMonths;
+    }
+
+    @Override
+    public void serialize(Enum<Month> value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+        if(_enableOneBaseMonths
+            /*
+                && we write the enum ordinal (int) value,
+                via WRITE_ENUMS_USING_INDEX or other similar construct
+            */
```
etrandafir93, line 28: @cowtowncoder - Hello again and sorry for the late reply, I finally got some time to continue working here.
I have applied your suggestions and refactored the code a bit. 

Regarding serialization, I am not sure how to check if we ar serializing the enum as an integer (via `WRITE_ENUMS_USING_INDEX` ) or other similar configuration. How should I do it?

cowtowncoder: Hmmh. Good question. Ideally we should be able to use `EnumSerializer` but that seems difficult if not impossible, at least by sub-classing (method `serialize()` is final for one but also initialization is quite tricky).
Similarly delegation is difficult to use wrt changing of value to serialize.

But `EnumSerializer` has useful pieces at least.

"createContextual":

https://github.com/FasterXML/jackson-databind/blob/2.17/src/main/java/com/fasterxml/jackson/databind/ser/std/EnumSerializer.java#L136C40-L136C40

"serialize()" : 

https://github.com/FasterXML/jackson-databind/blob/2.17/src/main/java/com/fasterxml/jackson/databind/ser/std/EnumSerializer.java#L168

and "_isShapeWrittenUsingIndex"

https://github.com/FasterXML/jackson-databind/blob/2.17/src/main/java/com/fasterxml/jackson/databind/ser/std/EnumSerializer.java#L263

implement logic.

cowtowncoder: And thinking out aloud further it'd be nice to use delegating serializer. There is `StdDelegatingSerializer` (but no intermediate `DelegatingSerializer` can't remember why not) which may or may not.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JavaTimeDeserializerModifier.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public JavaTimeDeserializerModifier(boolean oneBaseMonths) {
+        _oneBaseMonths = oneBaseMonths;
+    }
+
+    @Override
+    public JsonDeserializer<?> modifyEnumDeserializer(DeserializationConfig config, JavaType type, BeanDescription beanDesc, JsonDeserializer<?> defaultDeserializer) {
+        if (type.hasRawClass(Month.class)) {
+            return new MonthDeserializer((JsonDeserializer<Enum>) defaultDeserializer, _oneBaseMonths);
```
cowtowncoder, line 28: Shouldn't this only need to be returned if `_oneBasedMonths` is true? That might simplify handling a bit when there are fewer checks.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/JavaTimeSerializerModifier.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public JavaTimeSerializerModifier(boolean oneBaseMonths) {
+        _oneBaseMonths = oneBaseMonths;
+    }
+
+    @Override
+    public JsonSerializer<?> modifyEnumSerializer(SerializationConfig config, JavaType valueType, BeanDescription beanDesc, JsonSerializer<?> serializer) {
+        if (valueType.hasRawClass(Month.class)) {
+            return new MonthSerializer((JsonSerializer<Enum>) serializer, _oneBaseMonths);
```
cowtowncoder, line 28: Similar to deserializer: shouldn't this only need to be returned if `_oneBasedMonths` is true?
That might simplify handling a bit when there are fewer checks.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/JavaTimeSerializerModifier.java
```diff
@@ -0,0 +1,33 @@
...
+import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
+
+/**
+ * @since 2.17
+ */
+public class JavaTimeSerializerModifier extends BeanSerializerModifier {
+    private final boolean _oneBaseMonths;
+
+    public JavaTimeSerializerModifier() {
+        this(false);
```
cowtowncoder, line 18: Is this constructor needed?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JavaTimeDeserializerModifier.java
```diff
@@ -0,0 +1,32 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.deser.BeanDeserializerModifier;
+
+/**
+ * @since 2.17
+ */
+public class JavaTimeDeserializerModifier extends BeanDeserializerModifier {
+    private final boolean _oneBaseMonths;
+
+    public JavaTimeDeserializerModifier() {
```
cowtowncoder, line 17: Probably not needed?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JavaTimeDeserializerModifier.java
etrandafir93: @cowtowncoder - it's mostly about the naming and meaning of the classes. If we do it like this, perhaps it's better to rename the deserializer from `MonthDeserializer` to `OneBasedMonthDeserializer` and make it always updated the value returned by the delegate (without checking the `_oneBasedMonths ` flag -- and move the check here).
What do you think?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/JavaTimeSerializerModifier.java
etrandafir93: I removed it now 👍

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/JavaTimeSerializerModifier.java
etrandafir93: no, I removed it now 👍

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/MonthSerializer.java
etrandafir93: @cowtowncoder - I have tried a few things but couldn't find the best way of doing it using these methods. 
However, I managed to pass the WRITE_ENUMS_USING_INDEX feature as a Supplier<Boolean> that can be later checked , during the serialization - can you take a look?  si there a more straight-forward way of doing this?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JavaTimeDeserializerModifier.java
cowtowncoder: Yes, I think that makes sense -- I like explicit naming.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.java
```diff
@@ -142,6 +144,14 @@ public void setupModule(SetupContext context) {
         desers.addDeserializer(ZoneOffset.class, JSR310StringParsableDeserializer.ZONE_OFFSET);
 
         context.addDeserializers(desers);
+
+
+        boolean oneBasedMonth = _features.isEnabled(JavaTimeFeature.ONE_BASED_MONTHS);
+        Supplier<Boolean> serializeEnumsByIndex = () -> context.isEnabled(SerializationFeature.WRITE_ENUMS_USING_INDEX);
```
cowtowncoder, line 150: Let's not do this: setting should be accessed from either `SerializerProvider` or `SerializerConfig`: `SetupContext` is not guaranteed to be valid beyond time Modules are registered -- and adding a closure will keep a strong reference to it.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/OneBasedMonthSerializer.java
```diff
@@ -0,0 +1,32 @@
...
+    private final JsonSerializer<Enum> _defaultSerializer;
+
+    public OneBasedMonthSerializer(JsonSerializer<Enum> defaultSerializer, Supplier<Boolean> serializeEnumsByIndex) {
+        _defaultSerializer = defaultSerializer;
+        _serializeEnumsByIndex = serializeEnumsByIndex;
+    }
+
+    @Override
+    public void serialize(Enum<Month> value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+        if (_serializeEnumsByIndex.get()) {
```
cowtowncoder, line 25: Use `SerializerProvider.isEnabled(SerializationFeature.WRITE_ENUMS_USING_INDEX)` instead of passing `Supplier`.

cowtowncoder:
At this point, I'd need CLA before proceeding further.

cowtowncoder:
CLA received, not blocking any more.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/JavaTimeSerializerModifier.java
```diff
@@ -0,0 +1,32 @@
...
+import com.fasterxml.jackson.databind.ser.BeanSerializerModifier;
+
+/**
+ * @since 2.17
+ */
+public class JavaTimeSerializerModifier extends BeanSerializerModifier {
+    private final boolean _oneBaseMonths;
+    private final Supplier<Boolean> _serializeEnumsByIndex;
+
+    public JavaTimeSerializerModifier(boolean oneBaseMonths, Supplier<Boolean> serializeEnumsByIndex) {
```
cowtowncoder, line 19: As per other discussions, can drop `Supplier` and checks altogher as Serializer is only used if one-based-months feature is set.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/OneBasedMonthSerializer.java
cowtowncoder: Actually if/when serializer only registered when feature is enabled, can drop the check altogether.

cowtowncoder:
With CLA received, looks like there are only small changes needed:

1. Since "one-based" Feature check can be done by modifiers (or if we want, even by Module as modifiers only needed for 1-based ser/deser, for now), can drop checks by Serializer/Deserializer
2. Looks like some unit test(s) fail, need to resolve those.

and then I can merge this into 2.17 branch. Thanks!

cowtowncoder:
@etrandafir93  Ok: I fixed the one failing test (it was assuming coercion from Empty String is allowed by default; it is not for `Enum`s so changed test). WIll see if I can make the one last change to remove passing of `Supplier`.

cowtowncoder:
@etrandafir93 Ok, made all changes I wanted and ready to merge. But wanted to give a chance to see if there's anything you would want to change still. Please LMK and I'll merge this when ready!

etrandafir93:
@cowtowncoder - nothing else to add from my side.
Thank you for the guidance and support here!

cowtowncoder (approved):

PR #297, opened 2023-12-28
fix broken wiki link in README.md

pjfanning:


datatypes/README.md
```diff
@@ -73,4 +73,4 @@ System.out.println(withEmailJson);
 
 ## More
 
-See [Wiki](../../wiki) for more information (javadocs, downloads).
+See [Wiki](https://github.com/FasterXML/jackson-modules-java8/wiki) for more information (javadocs, downloads).
```
cowtowncoder, line 76: That works but relative would work too with one more `..` I think. Will merge, most important that it just works :)

cowtowncoder (approved):

PR #328, opened 2024-11-19
Add tests against #124

JooHyukKim:
verifies #124 behavior


cowtowncoder:
@JooHyukKim Great! We can either claim it's fixed in 2.18(.2) (since we don't know better), or, if you would like, change this against 2.17 and claim it's fixed 2.17(.4)?
Either way, if you could add an entry in `release-notes/VERSION-2.x` I'd be happy to merge PR with verification test & we can call it good!

JooHyukKim:
Oh I wasn't expecting 2.17 actually.
Let's merge this and do another backport PR for 2.17?

JooHyukKim:
> Either way, if you could add an entry in release-notes/VERSION-2.x I'd be happy to merge PR with verification test & we can call it good!

Will do!

JooHyukKim:
Closing in favor of #329

PR #335, opened 2024-12-07
Test #86 Cannot read `Optional`s written with `StdTypeResolverBuilder`

### LINKED ISSUE #86: Cannot read read written Optionals with StdTypeResolverBuilder
The following test fails on Jackson 2.9.7 and passes on 2.8.11:
```
    @Test
    public void test() throws Exception {
        ObjectMapper mapper = new ObjectMapper();
        mapper.setDefaultTyping(
                new StdTypeResolverBuilder()
                        .init(JsonTypeInfo.Id.CLASS, null)
                        .inclusion(JsonTypeInfo.As.WRAPPER_OBJECT)
        );
        mapper.registerModule(new Jdk8Module());
        String json = mapper.writeValueAsString(new Foo(Optional.of(1)));
        System.out.println(json);
        mapper.readValue(json, Foo.class);
    }

    public static class Foo {
        private Optional<Integer> bar;

        public Foo() {
        }

        public Foo(Optional<Integer> bar) {
            this.bar = bar;
        }

        public Optional<Integer> getBar() {
            return bar;
        }

        public void setBar(Optional<Integer> bar) {
            this.bar = bar;
        }
    }
```
Complete runnable project with this test-case can be found on https://github.com/isopov/jackson-optional-test

JooHyukKim:
- Add tests regarding #86 
- Blocked by https://github.com/FasterXML/jackson-databind/pull/4838

### DIFF
--- a/datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/OptionalDeserializer.java
+++ b/datatypes/src/main/java/com/fasterxml/jackson/datatype/jdk8/OptionalDeserializer.java
@@ -1,7 +1,10 @@
 package com.fasterxml.jackson.datatype.jdk8;
 
+import java.io.IOException;
 import java.util.Optional;
 
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.*;
 import com.fasterxml.jackson.databind.deser.ValueInstantiator;
 import com.fasterxml.jackson.databind.deser.std.ReferenceTypeDeserializer;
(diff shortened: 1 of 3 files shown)


cowtowncoder:
@JooHyukKim  Ok, this might be the way to do it, but we should actually change base (de)serializers from `jackson-databind` -- `ReferenceTypeSerializer` and `ReferenceTypeDeserializer`. That way `AtomicReference` would be covered, as would Scala `Option`. Possibly no change needed here, except for testing.
That would also simplify merging as master/3.0 has `Optional` directly supported by `jackson-databind` itself (2 out of 3 modules from here merged).

I also wonder if JSON serialization should (or should not) be changed for 3.0.
With 2.x we will basically omit `Optional` level altogether (both regular and polymorphic cases). But I think this was a mistake I made: I _think_ better way would have been to add level for polymorphic case to have:

```
{
  "java.util.Optional": {
      "com.fasterxml.jackson.datatype.jdk8.PolymorphicOptionalTest$Foo":{"bar":1}
  }
}
```

instead of what 2.x uses and expects:

```
{"com.fasterxml.jackson.datatype.jdk8.PolymorphicOptionalTest$Foo":{"bar":1}}
```

Obviously such change cannot be done for 2.x, but could be done for 3.x.
But even if change was made, we should probably add a `MapperFeature` for allowing enabling of old (2.x) behavior.

First things first however: this fix should be applied to `jackson-databind` base handlers.

JooHyukKim:
> but we should actually change base (de)serializers from `jackson-databind` -- `ReferenceTypeSerializer` and `ReferenceTypeDeserializer`

Yeah this makes sense. Let's work through `databind` module instead.

> I think better way would have been to add level for polymorphic case to have:

What benefit are we expecting out of the described serialization form?

JooHyukKim:
@cowtowncoder Filed fix PR in databind module https://github.com/FasterXML/jackson-databind/pull/4838 the PR contains how to test integrated etc...

cowtowncoder:
> > I think better way would have been to add level for polymorphic case to have:
> 
> What benefit are we expecting out of the described serialization form?

There may be multiple issues but the problem with `Optional` serialization wrt type info is that there are two things that could be polymorphic:

1. Contents of `Optional`. I think this is more about that and is not problematic (as long as `Optional` itself is known as type from enclosing class
2. Type of `Optional` itself, f.ex when nominal type is `java.lang.Object` and default typing is enabled. I think this would not work.

But I must admit that the interaction gets pretty complicated and fundamentally it might be that it is not possibly to only add new level for polymorphic case.
In fact original "unwrapped" approach to `Optional` serialization might be a bad decision form my part. So, for example, for `Optional<Integer>(5)` is now serialized simply as

```
5
```

and `Optional<Integer>(null)` (or rather `Optional.empty()`) as

```
null
```

but it 'd have been simpler, in the end if those instead were serialized as

```
[ 5 ]
```

and 

```
[ ]
```

because then we wouldn't need convert `null` into `Optional.empty()` (plus actually be able to distinguish `null` and `Optional.empty()` when deserializing).

But it may be bit late to change this even for 3.0. And there hasn't been too many reports for problems (there are some, wrt nested `Optional`s but not a ton.

So let's focus on immediate issue and forget about what I was proposing for now. :)

JooHyukKim:
Closing due to reasons I mentioned in [comment](https://github.com/FasterXML/jackson-databind/pull/4838#issuecomment-2558684339)

PR #343, opened 2025-01-22
Time zone aware lenient parsing

### LINKED ISSUE #342: Lenient deserialization of `LocalDate`, `LocalDateTime` is not time-zone aware
The deserialization code in `LocalDateDeserializer` contains lenient logic to also accept strings like `2025-01-21T23:00:00Z` and deserialize this to a `LocalDate`: https://github.com/FasterXML/jackson-modules-java8/blob/2.19/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserializer.java#L164

The problem is that this example timestamp will always be deserialized to `2025-01-21`, although in time zones with a positive offset (e.g. Central European Time = +01:00), this is unexpected and a source of errors for users.

My use case is also a JavaScript client communicating with a Java backend (as motivated in one of the comments in the code) and I appreciate the idea. But I think the conversion should be time-zone aware. Strings ending with `Z` could e.g. be parsed as `Instant`s which could then be transformed to a `LocalDate` according to the JVM's default `TimeZone` or the `TimeZone` that has been set on the `ObjectMapper`.

hpoettker:
Resolves #342.

The current implementation of lenient parsing of `LocalDate` and `LocalDateTime` was implemented with the goal to always produce the values such that they need to be interpreted to be UTC.

In particular for `LocalDate` this can very easily lead to confusing behavior for users.

The proposed change only leads to a change in behavior when the `TimeZone` is explicitly set on the `ObjectMapper` and is different from `UTC`.

### DIFF
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
@@ -9,6 +9,7 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.Temporal;
 import java.util.Map;
+import java.util.TimeZone;
 
 import com.fasterxml.jackson.annotation.OptBoolean;
 import com.fasterxml.jackson.databind.cfg.CoercionAction;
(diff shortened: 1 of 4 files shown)


datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/LocalDateDeserTest.java
```diff
@@ -153,6 +154,15 @@ public void testDeserializationAsString03() throws Exception
                 value);
     }
 
+    @Test
+    public void testDeserializationAsString04() throws Exception
+    {
+        ObjectReader reader = READER.with(TimeZone.getTimeZone(Z_BUDAPEST));
+        Instant instant = Instant.parse("2024-07-21T22:00:00Z");
+        LocalDate value = reader.readValue('"' + instant.toString() + '"');
+        assertEquals("The value is not correct.", LocalDate.parse("2024-07-22"), value);
```
cowtowncoder, line 163: I think this is wrong and against the semantics of `LocalDate`.

cowtowncoder:
I'll continue discussion on #342 but basically I do not agree with the change as is.

Mis-using `LocalDate` to take in zoned-values may have been a mistake all along, but I don't want to compound be making it even more ill-conceived.

In this case you will probably need to register custom deserializer to solve your use case.

hpoettker:
I understand your objections. Then let's close this one.

Issue #339, opened 2025-01-11
JSTEP-10: Unifying testing structure/tools 


JooHyukKim:
This issue serves as an umbrella issue for resolving [[JSTEP-10: Unifying testing structure/tools]](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-10)

### TODO

- Add JUnit 5 dep + `failing` test utils
- Migrate modules to JUnit 5
- Migrate failing tests to `/tofix` and remove existing `/failing` funciontality
- Remove JUnit 4

Issue #207, opened 2021-03-21
Fail to serialize `TemporalAdjuster` type with 2.12


gokhanoner:
Starting with 2.12, below code fails & throw exception. I wonder if it's related to https://github.com/FasterXML/jackson-databind/issues/2683

Tested with 2.11.4 & 2.10.5, both serialize successfully.

Note: Adding deprecated jsr310 module also doesn't work.

```
public class JacksonTest {
    private static ObjectMapper objectMapper;

    @BeforeClass
    public static void setup() {
        objectMapper = JsonMapper.builder()
                .addModules(
                        new JavaTimeModule()
                )
                .build();
    }

    @Test
    public void whenUsingTemporalAdjusterExpectTrue() throws JsonProcessingException {
        TestPojo mrd = new TestPojo();
        System.out.println(objectMapper.writeValueAsString(mrd));
    }

    static class TestPojo {
        private TemporalAdjuster ta;

        TestPojo() {
            this.ta = TemporalAdjusters.firstDayOfMonth();
        }

        public TemporalAdjuster getTa() {
            return ta;
        }

        public void setTa(TemporalAdjuster ta) {
            this.ta = ta;
        }
    }
}

```

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.temporal.TemporalAdjusters$$Lambda$80/0x0000000800c07840` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: com.oner.JacksonTest$TestPojo["ta"])

	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:77)
	at com.fasterxml.jackson.databind.SerializerProvider.reportBadDefinition(SerializerProvider.java:1276)
	at com.fasterxml.jackson.databind.ser.impl.UnsupportedTypeSerializer.serialize(UnsupportedTypeSerializer.java:35)
	at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:728)
	at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:770)
	at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:178)
[...]

PR #348, opened 2025-01-27
[JSTEP-10] Migrate remaining `datetime` module tests to JUnit 5

JooHyukKim:
part of #339
blocked by #347


cowtowncoder (approved):

cowtowncoder:
Proceeding with this because I started the process. Will tackle that one failing unit test separately.

cowtowncoder:
@JooHyukKim Figured out most fails: there were cases where `assumeTrue` / `assumeFalse` were converted to assertions -- this is not correct as Assumptions behave different from Assertions.

There's also one other issue, wrt #207, should be able to resolve that too but first things first.

PR #350, opened 2025-01-27
Fix 337 : Negative `Duration`s are wrongly calculated

### LINKED ISSUE #337: Negative `Duration` does not round-trip properly with `WRITE_DURATIONS_AS_TIMESTAMPS` enabled
I found this similar issue https://github.com/FasterXML/jackson-modules-java8/issues/165 but it seems different, as that user reported a bug on `2.10.2` and refers to a different config option.

`WRITE_DURATIONS_AS_TIMESTAMPS` defaults to true.

### Before v2.12.0
`WRITE_DURATIONS_AS_TIMESTAMPS` set to true.

It seems that a negative value...
* Serializes incorrectly ❌ 
* Deserializes back to the original value ✅ 

### v2.12.0 and later (including 2.18.2)
`WRITE_DURATIONS_AS_TIMESTAMPS` set to true.

It seems that a negative value...
* Serializes correctly ✅ 
* Deserializes incorrectly ❌ 

Here is an example repo demonstrating the problem https://github.com/jmuia/jackson-duration-serialization

JooHyukKim:
fixes #337


JooHyukKim:
@cowtowncoder Oh okay so TimeAdjuster test issue was fixed #352!
Will get back to this again. Thanks @cowtowncoder !

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/DecimalUtils.java
```diff
@@ -115,12 +115,24 @@ public static int extractNanosecondDecimal(BigDecimal value, long integer)
      * on some JRE releases.
      *
      * @since 2.9.8
+     * @deprecated Use {@link #extractSecondsAndNanos(BigDecimal, BiFunction, boolean)} instead.
```
cowtowncoder, line 118: Should add "since 2.19" here, will add.

PR #354, opened 2025-02-03
Add feature for time-zone aware lenient date parsing

### LINKED ISSUE #342: Lenient deserialization of `LocalDate`, `LocalDateTime` is not time-zone aware

hpoettker:
Resolves #342

As proposed in the discussion of the issue, the PR adds a value in `JavaTimeFeature` that enables the time-zone to be used when leniently parsing `LocalDate` and `LocalDateTime` from the UTC/ISO instant format.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -20,6 +20,19 @@ public enum JavaTimeFeature implements JacksonFeature
     NORMALIZE_DESERIALIZED_ZONE_ID(true),
 
     /**
+     * Feature that determines whether the {@link java.util.TimeZone} of the
+     * {@link com.fasterxml.jackson.databind.DeserializationContext} is used
+     * when leniently deserializing {@link java.time.LocalDate} or
+     * {@link java.time.LocalDateTime} from the UTC/ISO instant format.
+     * <p>
+     * Default setting is disabled, for backwards-compatibility with
+     * Jackson 2.18.
+     *
+     * @since 2.19
+     */
+    USE_TIME_ZONE_FOR_LENIENT_DATE_PARSING(false),
+
+    /**
      * Feature that controls whether stringified numbers (Strings that without
      * quotes would be legal JSON Numbers) may be interpreted as
      * timestamps (enabled) or not (disabled), in case where there is an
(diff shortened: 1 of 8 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
```diff
@@ -19,6 +19,17 @@ public enum JavaTimeFeature implements JacksonFeature
...
 
+    /**
+     * Feature that determines whether the {@link java.util.TimeZone} of the
+     * {@link com.fasterxml.jackson.databind.DeserializationContext} is used
+     * when leniently deserializing {@link java.time.LocalDate} or
+     * {@link java.time.LocalDateTime} from the UTC/ISO instant format.
+     * <p>
+     * Default setting is disabled, for backwards-compatibility with
+     * Jackson 2.18.
+     */
```
cowtowncoder, line 30: please add `@since 2.19` javadoc tag

cowtowncoder (approved):
LGTM!

cowtowncoder:
@hpoettker Thank you for the pr -- it looks really good! Before I can merge there's just one formality (if not done before?) -- getting CLA. Needs to be sent just once, before the first contribution. 

Doc is here: https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf

and the usual way is to print, fill  & sign, scan/photo, email to `cla` at fasterxml dot com.
Once receive it CLA is good for all future contributions and I can proceed with merging!

Looking forward to getting this merged.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
hpoettker: done

hpoettker:
Thanks for the timely review!

It's my first contribution, and I've just now sent the signed CLA via mail.

Issue #291, opened 2023-11-30
`InstantDeserializer` fails to parse negative numeric timestamp strings for pre-1970 values


advorako:
Encountered while debugging through a project making use of this for Avro serialization/deserialization:

For an object containing a java.time.Instant with associated Avro schema field of "type" : [ "null", { "type" : "string", "java-class" : "java.time.Instant"} ] ...
- A non-null Instant value will serialize into a numeric timestamp string, which combines its epoch seconds and nanoseconds offset member values, separated by a period '.' character.
- The epoch seconds portion may be positive or negative, depending on which side of the Unix Epoch (1970-01-01) it lands on.
- For example, January 1, 1970 at 00:00:03 becomes "3.000000000", while December 31, 1969 at 23:59:57 becomes "-3.000000000".

InstantDeserializer._fromString() is used to try and parse these numeric timestamp strings, but its _countPeriods() helper method does not account for the possible leading '-' character
https://github.com/FasterXML/jackson-modules-java8/blob/2.16/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java#L369-L384

This causes _fromString() to skip over the parsing method that would work in this situation ... 
https://github.com/FasterXML/jackson-modules-java8/blob/2.16/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java#L409-L410
... and instead fall through into DateTimeFormatter, which cannot handle this type of string and throws a DateTimeParseException
https://github.com/FasterXML/jackson-modules-java8/blob/2.16/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java#L433

Example code:

```
public class Example
{
    public static class InstantWrapper
    {
        public Instant time;
        public InstantWrapper() { time = null; }
        public InstantWrapper(final Instant time) { this.time = time; }
    }

    public static void main(final String[] args) throws IOException
    {
        final String instantWrapperSchema = "{\n"
            +"\"type\": \"record\",\n"
            +"\"name\": \"InstantWrapper\",\n"
            +"\"fields\": [\n"
            +"{\"name\": \"time\", \"type\" : [ \"null\", {\"type\" : \"string\", \"java-class\" : \"java.time.Instant\"} ]}\n"
            +"]}";

        final AvroSchema avroSchema = new AvroSchema(new Schema.Parser().setValidate(true).parse(instantWrapperSchema));

[...]

PR #362, opened 2025-03-03
Fix(#291): InstantDeserializer fails to parse negative numeric timestamp strings for pre-1970 values

Strongbeard:
Fix for #291:

- Updated `InstantDeserializer._countPeriods(String)` to check for `-` and `+` at beginning of numeric string.
- Additionally, added check if `ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS` is enabled. If disabled, throws same exception as before change from `_countPeriods(...)`.
- Added associated tests to `failing` directory and then moved to `deser` directory.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -382,12 +382,12 @@ public class InstantDeserializer<T extends Temporal>
     }
 
     // Helper method to find Strings of form "all digits" and "digits-comma-digits"
-    protected int _countPeriods(JsonParser p, String str)
+    protected int _countPeriods(String str, boolean allowLeadingPlusSign)
     {
         int commas = 0;
         int i = 0;
         int ch = str.charAt(i);
-        if (ch == '-' || (ch == '+' && p.isEnabled(ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS.mappedFeature()))) {
+        if (ch == '-' || (ch == '+' && allowLeadingPlusSign)) {
             ++i;
         }
         for (int end = str.length(); i < end; ++i) {
@@ -420,7 +420,7 @@ public class InstantDeserializer<T extends Temporal>
                 _formatter == DateTimeFormatter.ISO_ZONED_DATE_TIME
             ) {
             // 22-Jan-2016, [datatype-jsr310#16]: Allow quoted numbers too
-            int dots = _countPeriods(p, string);
+            int dots = _countPeriods(string, p.isEnabled(ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS.mappedFeature()));
             if (dots >= 0) { // negative if not simple number
                 try {
                     if (dots == 0) {
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
@@ -44,6 +44,8 @@ import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static com.fasterxml.jackson.core.json.JsonReadFeature.ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS;
+
 /**
  * Deserializer for Java 8 temporal {@link Instant}s, {@link OffsetDateTime},
  * and {@link ZonedDateTime}s.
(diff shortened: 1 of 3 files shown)


JooHyukKim (approved):
Looks good

Strongbeard:
Would it be better to explicitly throw a `JsonParseException` when a `+` is encountered and `p.isEnabled(ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS.mappedFeature())` is false? I didn't think about it when writing the PR, but a more straightforward error message might be better for users...

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -380,11 +382,16 @@ protected boolean shouldReadTimestampsAsNanoseconds(DeserializationContext conte
     }
 
     // Helper method to find Strings of form "all digits" and "digits-comma-digits"
-    protected int _countPeriods(String str)
+    protected int _countPeriods(JsonParser p, String str)
```
cowtowncoder, line 385: Instead of passing `JsonParser`, pass the feature as `boolean`.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
```diff
@@ -380,11 +382,16 @@ protected boolean shouldReadTimestampsAsNanoseconds(DeserializationContext conte
...
     {
         int commas = 0;
-        for (int i = 0, end = str.length(); i < end; ++i) {
-            int ch = str.charAt(i);
+        int i = 0;
+        int ch = str.charAt(i);
+        if (ch == '-' || (ch == '+' && p.isEnabled(ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS.mappedFeature()))) {
```
cowtowncoder, line 390: Aside from moving access to feature flag to caller, avoid use of `mappedFeature()` if possible? (if there is `JsonParser.isEnabled(JsonReadFeature)` -- can't remember if there was)

cowtowncoder:
> Would it be better to explicitly throw a `JsonParseException` when a `+` is encountered and `p.isEnabled(ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS.mappedFeature())` is false? I didn't think about it when writing the PR, but a more straightforward error message might be better for users...

I think this is ok. Parsing numbers from Strings is bit icky anyway.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
Strongbeard: Will do.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
Strongbeard: While I could pass `JsonParser.Feature.ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS` directly instead of indirectly via the call to `.mappedFeature()`, it is marked `@Deprecated` in the `jackson-core` `2.18` release branch and its comments directly instruct one to use `JsonReadFeature.ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS` instead. It appeared to me that the intention was to migrate away from the deprecated enums to this new set instead and that `.mappedFeature()` provided temporary backward compatibility until `JsonParser.isEnabled(JsonReadFeature)` is added to the API to replace `JsonParser.isEnabled(JsonParser.Feature)`. Hence why `@SuppressWarnings("deprecation")` is used so much throughout `JsonReadFeature`. Did I interpret the intention behind that deprecation incorrectly?

The only other approach I see that is not marked deprecated would be to deal with the lower-level bit masks that turn the features on/off directly, but I get the sense that bypassing the higher level abstraction of the different Feature interfaces would be counter-productive to the overall direction of this project.

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
Strongbeard: Implemented in commit aa6f70a1b2282621e4e1501363ce1389b90f0020

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/InstantDeserializer.java
Strongbeard: I ran the following `grep` across `jackson-core`, `jackson-databind`, and `jackson-modules-java8` all on release branch `2.18`, but could not get a hit that did not immediately call `.mappedFeature()` to pass to the `isEnabled(Feature)` function. `JsonMapper` in `jackson-databind` has a function signature `isEnabled(JsonReadFeature)`, but it calls `.mappedFeature()` inside it.
```
grep -r --include='*.java' --exclude-dir='bin' --exclude-dir='target' --exclude-dir='build' 'nabled(JsonReadFeature'
```

cowtowncoder: @Strongbeard sorry, my bad. TL;DNR: PR is fine as-is. :)

This is one area where 2.x refactoring (splitting of `JsonParser.Feature` into `StreamReadFeature` / `JsonReadFeature`) is ugly -- there is no `JsonParser.isEnabled(JsonReadFeature)`.

And in fact... this is bit tricky since code will not port into 3.0 correctly as JSON-specific feature won't be available any more.
But I think I'll handle that when merging by assuming leading plus sign is allowed on Stringified-numbers.

Strongbeard: No problem. Thanks for volunteering to handle porting it to 3.0.

I submitted my signed Contributor Agreement via email yesterday, so this should be ready on my end for whenever the maintainers wish to merge it.

cowtowncoder (approved):

PR #363, opened 2025-03-13
Fix #279 `OffsetDateTime` deserialization fail with date-time pattern 

### LINKED ISSUE #279: `OffsetDateTime`deserialization fails when using (specific?) date-time pattern
If we have an OffsetDateTime property inside a DTO  where we specify the pattern, the deserialization operation fails: 
Ex: 
Inside OffsetDateDto: 

```
   private OffsetDateTime date= OffsetDateTime.now(ZoneId.of("UTC")).truncatedTo(ChronoUnit.SECONDS);

    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'")
    public OffsetDateTime getDate() {
        return date;
    }
    public void setDate(OffsetDateTime date) {
        this.date = date;
    }
```

The code of the application: 

```
           ObjectMapper o = new ObjectMapper();
            o.findAndRegisterModules();
            OffsetDateDto dto = new OffsetDateDto();
            StringWriter w = new StringWriter();
            o.writeValue(w, dto);
            String s = w.toString();
            System.out.println(s);
            OffsetDateDto x = o.readValue(s.getBytes(StandardCharsets.UTF_8), OffsetDateDto.class);
```

When we write the value (serialize dto I get displayed: '{"date":"2023-09-01T11:07:48Z",...'}
But  the last line throws an error because deserialization fails: 

```
com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.OffsetDateTime` from String "2023-09-01T11:07:48Z": Failed to deserialize java.time.OffsetDateTime: (java.time.DateTimeException) Unable to obtain OffsetDateTime from TemporalAccessor: {},ISO resolved to 2023-09-01T11:07:48 of type java.time.format.Parsed
 at [Source: (byte[])"{"date":"2023-09-01T11:07:48Z","localDate":"2023-09-01","localDateTime":"2023-09-01T11:07:48.7458067"}"; line: 1, column: 9] (through reference chain: json.OffsetDateDto["date"])
	at com.fasterxml.jackson.databind.exc.InvalidFormatException.from(InvalidFormatException.java:67)
	at com.fasterxml.jackson.databind.DeserializationContext.weirdStringException(DeserializationContext.java:1991)
	at com.fasterxml.jackson.databind.DeserializationContext.handleWeirdStringValue(DeserializationContext.java:1219)
	at com.fasterxml.jackson.datatype.jsr310.deser.JSR310DeserializerBase._handleDateTimeException(JSR310DeserializerBase.java:176)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer._fromString(InstantDeserializer.java:307)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.deserialize(InstantDeserializer.java:217)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer.deserialize(InstantDeserializer.java:51)
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:313)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:176)
	at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:322)
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4674)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3690)
	at json.JsonObjectMapperTester.testOffsetDate(JsonObjectMapperTester.java:60)
	at a.Starter.main(Starter.java:21)
Caused by: java.time.DateTimeException: Unable to obtain OffsetDateTime from TemporalAccessor: {},ISO resolved to 2023-09-01T11:07:48 of type java.time.format.Parsed
	at java.base/java.time.OffsetDateTime.from(OffsetDateTime.java:372)
	at com.fasterxml.jackson.datatype.jsr310.deser.InstantDeserializer._fromString(InstantDeserializer.java:302)
	... 10 more
Caused by: java.time.DateTimeException: Unable to obtain ZoneOffset from TemporalAccessor: {},ISO resolved to 2023-09-01T11:07:48 of type java.time.format.Parsed
	at java.base/java.time.ZoneOffset.from(ZoneOffset.java:350)
	at java.base/java.time.OffsetDateTime.from(OffsetDateTime.java:361)
	... 11 more
```

I've tried also with the pattern 'yyyy-MM-dd'T'HH:mm:ssX' and I get the same error.
Am I doing something wrong here? 

Thanks in advance

JooHyukKim:
resolves #279 I am not sure about clean-ness of the solution.
Open for modification!

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
@@ -1,6 +1,7 @@
 package com.fasterxml.jackson.datatype.jsr310.deser;
 
 import java.io.IOException;
+import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.ResolverStyle;
@@ -168,6 +169,11 @@ public abstract class JSR310DateTimeDeserializerBase<T>
             //a time zone picked up from JsonFormat annotation, otherwise serialization might not work
             if (formatOverrides.hasTimeZone()) {
                 df = df.withZone(formatOverrides.getTimeZone().toZoneId());
+            } else {
+                // [#279] OffsetDateTimedeserialization fails when using (specific?) date-time pattern #279
+                if (pattern.endsWith("'Z'")) {
+                    df = df.withZone(ZoneOffset.UTC);
+                }
             }
             deser = deser.withDateFormat(df);
         }
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/OffsetDateTimeDeser279Test.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/OffsetDateTimeDeser279Test.java
@@ -22,7 +22,7 @@ public class OffsetDateTimeDeser279Test
         public Wrapper279(OffsetDateTime d) { date = d; }
         protected Wrapper279() { }
 
-        @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'")
+        @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'") //
         public OffsetDateTime getDate() {
             return date;
         }
--- a/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/OffsetDateTimeDeser279Test.java
+++ b/datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/OffsetDateTimeDeser279Test.java
@@ -1,4 +1,4 @@
-package com.fasterxml.jackson.datatype.jsr310.tofix;
+package com.fasterxml.jackson.datatype.jsr310.deser;
 
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
@@ -9,13 +9,13 @@ import org.junit.jupiter.api.Test;
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.ModuleTestBase;
-import com.fasterxml.jackson.datatype.jsr310.testutil.failure.JacksonTestFailureExpected;
 
 import static org.junit.jupiter.api.Assertions.*;
 
-public class OffsetDateTimeDeser279Test extends ModuleTestBase
+// [modules-java8#279] OffsetDateTimeDeserializer fails to parse date-time string with 'Z' at the end
+public class OffsetDateTimeDeser279Test
+    extends ModuleTestBase
 {
-    // For [modules-java8#279]
     static class Wrapper279 {
         OffsetDateTime date;
 
@@ -31,18 +31,22 @@ public class OffsetDateTimeDeser279Test extends ModuleTestBase
         }
     }
 
-    private ObjectMapper MAPPER = newMapper();
+    private final ObjectMapper MAPPER = newMapper();
 
-    // For [modules-java8#279]
-    @JacksonTestFailureExpected
     @Test
-    public void testWrapperWithPattern279() throws Exception
+    public void testWrapperWithPattern279()
+        throws Exception
     {
-        final OffsetDateTime date = OffsetDateTime.now(ZoneId.of("UTC"))
+        OffsetDateTime date = OffsetDateTime.now(ZoneId.of("UTC"))
+                .withYear(2025).withMonth(3).withDayOfMonth(13).withHour(13).withMinute(29).withSecond(57).withNano(0)
                 .truncatedTo(ChronoUnit.SECONDS);
-        final Wrapper279 input = new Wrapper279(date);
-        final String json = MAPPER.writeValueAsString(input);
+        Wrapper279 input = new Wrapper279(date);
+        
+        // serialization first
+        String json = MAPPER.writeValueAsString(input);
+        assertEquals("{\"date\":\"2025-03-13T13:29:57Z\"}", json);
 
+        // deserialization second
         Wrapper279 result = MAPPER.readValue(json, Wrapper279.class);
         assertEquals(input.date, result.date);
     }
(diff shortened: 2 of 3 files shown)


datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/OffsetDateTimeDeser279Test.java
```diff
@@ -9,20 +9,20 @@
...
 {
-    // For [modules-java8#279]
     static class Wrapper279 {
         OffsetDateTime date;
 
         public Wrapper279(OffsetDateTime d) { date = d; }
         protected Wrapper279() { }
 
-        @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'")
+        @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss'Z'") //
```
JooHyukKim, line 25: Any suggestions on patterns to test?

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/JSR310DateTimeDeserializerBase.java
```diff
@@ -168,6 +169,11 @@ protected JSR310DateTimeDeserializerBase<?> _withFormatOverrides(Deserialization
             //a time zone picked up from JsonFormat annotation, otherwise serialization might not work
             if (formatOverrides.hasTimeZone()) {
                 df = df.withZone(formatOverrides.getTimeZone().toZoneId());
+            } else {
+                // [#279] OffsetDateTimedeserialization fails when using (specific?) date-time pattern #279
+                if (pattern.endsWith("'Z'")) {
+                    df = df.withZone(ZoneOffset.UTC);
+                }
```
JooHyukKim, line 176: Super single-case solution... may be improve handling along the way

datetime/src/test/java/com/fasterxml/jackson/datatype/jsr310/deser/OffsetDateTimeDeser279Test.java
cowtowncoder: I can't believe I suggest this but... my hesitation on considering quoted 'Z' as anything other than constant character to include (and NOT have any semantic effect) makes me wonder if someone should just ask ChatGPT (or LLM-based code assistants) true meaning of the pattern here, its semantics?

It could be the original bug report itself might be wrong.

Consider `T` here: that is hard-coded separator and not taken to be semantic part of date/time value. Similarly should pattern really be

    yyyy-MM-dd'T'HH:mm:ssZ

instead? Or, if not, shouldn't affect processing of underlying date/time value in any way (just included as effectively decoration)

JooHyukKim: Answer from GPT below

JooHyukKim: In Java's date and time formatting, single quotes (') are used to denote literal characters within a pattern string. This means that any text enclosed in single quotes is treated as-is and does not have any special meaning in terms of date or time components.

Understanding the 'Z' Character:

Without Single Quotes (Z): In date-time formatting patterns, an unquoted Z typically represents the RFC 822 time zone, which corresponds to the time zone offset from UTC. For example, in the pattern "yyyy-MM-dd'T'HH:mm:ssZ", the Z at the end expects a numeric time zone offset like +0800 or -0500.

With Single Quotes ('Z'): When Z is enclosed in single quotes, as in "yyyy-MM-dd'T'HH:mm:ss'Z'", it is treated as a literal character. This means the formatter expects the character Z to appear exactly at that position in the date-time string, without interpreting it as a time zone indicator.

Practical Implications:

Parsing: If you have a date-time string like "2025-03-13T13:29:57Z", where Z indicates UTC time (also known as "Zulu" time), using the pattern "yyyy-MM-dd'T'HH:mm:ss'Z'" will treat the Z as a literal character. This means the parser will expect the Z in the string but won't interpret it as indicating the UTC time zone. As a result, the parsed date-time may not be correctly set to UTC.

Formatting: When formatting a date-time object to a string, using the pattern "yyyy-MM-dd'T'HH:mm:ss'Z'" will produce a string with a literal Z at the end, regardless of the actual time zone of the date-time object. For example, it might produce "2025-03-13T13:29:57Z", even if the date-time is not in UTC.

Recommendation:

To correctly parse and format date-time strings that include a Z to indicate UTC time, it's advisable to use patterns that interpret the Z appropriately. For ISO 8601 date-times, the X pattern letter is suitable:

Pattern: "yyyy-MM-dd'T'HH:mm:ssX"
This pattern handles ISO 8601 time zone indicators correctly, interpreting Z as UTC and other offsets like +02:00 appropriately.

Example:

java
Copy
Edit
String dateTimeString = "2025-03-13T13:29:57Z";
DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssX");
OffsetDateTime dateTime = OffsetDateTime.parse(dateTimeString, formatter);
In this example, the OffsetDateTime object dateTime will correctly represent the UTC date and time specified in the string.

For more details on date-time formatting patterns, you can refer to the [SimpleDateFormat documentation](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html).


Sources

JooHyukKim: Re-reading the original issue makes me think that we need to organize things into a list of what input/output we expect

cowtowncoder: ChatGPT gave seemingly extensive and also (yay!) accurate answer.

cowtowncoder:
@JooHyukKim Which version do we think fix went in? 2.19.0? I don't see an entry in release notes (https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.19) or `release-notes/VERSION-2.x`.

JooHyukKim:
[Conclusion was working as expected](https://github.com/FasterXML/jackson-modules-java8/issues/279#issuecomment-2735285533)

PR #365, opened 2025-03-26
Fix #364: make one-based Month deserializer accept "12"

### LINKED ISSUE #364: Deserialization of Month in ONE_BASED_MONTHS mode fails for value "12"
When `JavaTimeFeature.ONE_BASED_MONTHS` is in effect, I expect to convert strings `"1"` to `Month.JANUARY` ... `"12"` to `Month.DECEMBER`. For values 1-11 it works, but `"12"` raises error.

The cause is that `com.fasterxml.jackson.datatype.jsr310.deser.OneBasedMonthDeserializer` first invokes the original, 0-based deserializer and then it tweaks its output with `zeroBaseMonth.minus(1)`. The original deserializer can handle values `"0"` to `"11"`, but `"12"` is obviously out of range.

bbobcik:
As described in #364 , `com.fasterxml.jackson.datatype.jsr310.deser.OneBasedMonthDeserializer` incorrectly assumed that all values can be converted to a `Month` value by the original deserializer. This assumptions fails for `"12"`.

In this fix the `OBMD` is refactored in order to depend less on the original deserializer. Certain inputs (1-digit and 2-digit numbers and their string representations) are handled - validated and parsed - directly. Non-numeric inputs as well as numbers that have 3 or more digits are forwarded to the delegate, as before.

Associated tests are expanded to cover all months.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/OneBasedMonthDeserializer.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/OneBasedMonthDeserializer.java
@@ -23,20 +23,19 @@ public class OneBasedMonthDeserializer extends DelegatingDeserializer {
     @Override
     public Object deserialize(JsonParser parser, DeserializationContext context) throws IOException {
         JsonToken token = parser.currentToken();
-        if (_isPossibleNumericValue(token)) {
-            String monthSpec = parser.getText();
-            int oneBasedMonthNumber = _decodeNumber(monthSpec);
-            if (1 <= oneBasedMonthNumber && oneBasedMonthNumber <= 12) {
-                return Month.of(oneBasedMonthNumber);
-            } else if (oneBasedMonthNumber >= 0) {
-                throw new InvalidFormatException(parser, "Month number " + oneBasedMonthNumber + " not allowed for 1-based Month.", oneBasedMonthNumber, Integer.class);
-            }
+        switch (token) {
+            case VALUE_NUMBER_INT:
+                return _decodeMonth(parser.getIntValue(), parser);
+            case VALUE_STRING:
+                String monthSpec = parser.getText();
+                int oneBasedMonthNumber = _decodeNumber(monthSpec);
+                if (oneBasedMonthNumber >= 0) {
+                    return _decodeMonth(oneBasedMonthNumber, parser);
+                }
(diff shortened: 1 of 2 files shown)


datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/OneBasedMonthDeserializer.java
```diff
@@ -17,31 +16,51 @@
...
     public Object deserialize(JsonParser parser, DeserializationContext context) throws IOException {
         JsonToken token = parser.currentToken();
-        Month zeroBaseMonth = (Month) getDelegatee().deserialize(parser, context);
-        if (!_isNumericValue(parser.getText(), token)) {
-            return zeroBaseMonth;
-        }
-        if (zeroBaseMonth == Month.JANUARY) {
-            throw new InvalidFormatException(parser, "Month.JANUARY value not allowed for 1-based Month.", zeroBaseMonth, Month.class);
+        if (_isPossibleNumericValue(token)) {
+            String monthSpec = parser.getText();
```
cowtowncoder, line 27: Ideally this should not use `getText()` if we get actual number (`JsonToken.VALUE_NUMBER_INT`) but just call `parser.getIntValue()`.

cowtowncoder:
First of all: thank you for reporting the issue and providing a fix! It looks good overall.

But one thing we need before merging (if not done earlier): we'd need CLA (from here: https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf ).
It is needed just once before merging the first PR; works for all future contributions (so if you have sent one in the past please LMK).

The usual way is to print it, fill & sign, email to `cla` at fasterxml dot com.

Once we have it, we can proceed with merging.

Thanks again!

datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/deser/OneBasedMonthDeserializer.java
bbobcik: Agreed, reworked in https://github.com/FasterXML/jackson-modules-java8/pull/365/commits/3f564fa04cb3677604bce0edb8aa146f4740e37a

cowtowncoder:
CLA received: hoping to review and merge this weekend.

cowtowncoder (approved):

PR #381, opened 2025-12-29
Fix a potential problem in `JavaTimeModule._findFactory()`

AlbertLovers:
Validate the order of the argument types against the order of the parameters. The previous version ignored mismatches and returned the first method with the correct name and correct number of parameters regardless of parameter types.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.java
@@ -270,15 +270,23 @@ public final class JavaTimeModule
                     || (method.getParameterCount() != argCount)) {
                 continue;
             }
-            for (int i = 0; i < argCount; ++i) {
-                Class<?> argType = method.getParameter(i).getRawType();
-                if (!argType.isAssignableFrom(argTypes[i])) {
-                    continue;
-                }
+            if (!allArgTypesMatch(argTypes, method)) {
+                continue;
             }
             return method;
         }
         return null;
+    }
+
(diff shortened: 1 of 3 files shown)


cowtowncoder:
First of all, thank you for contributing this fix!

Second: one (and only) process thing we have is that we need CLA from:

https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf

before the first PR (but only needs to be sent once, good for all future contributions).
The easiest way is usually to print it, fill & sign, email to `cla` at fasterxml dot com.
Once I get it, I can merge the PR.

Third: is there specific problem case this solves? If so, would it be possible to add a unit test to guard against regression?

AlbertLovers:
The contributor agreement is on its way.

> Third: is there specific problem case this solves? If so, would it be possible to add a unit test to guard against regression?

I did not run into a specific issue with the code, I discovered the issue while looking through the `JavaTimeModule` while working on a migration. It seems to me the `for` loop checking the types is either important but broken (which makes me wonder why there are no bug reports) or it is superfluous and needs to be removed.

Unfortunately the `JavaTimeModule` does not seem to have unit tests yet and I am not familiar enough with the workings of `AnnotatedClass` and `AnnotatedMethod` to write meaningful tests.

I can write tests for the `_findFactory` method if you don't mind me introducing something like `Mockito` to the testsuite.

cowtowncoder:
@AlbertLovers Ah ok. Looking at code it's called just once,a s 

    AnnotatedMethod factory = _findFactory(ac, "of", String.class);

for `ZoneId`.

Given this, I think it's ok not to write mock-based test; some indirect testing probably exists wrt deserialization of JDK `ZoneId` instances.

Fix itself makes sense.

cowtowncoder:
Actually... no, looks like this code never gets executed. Oh well. Even less point in writing mock tests.

Issue #299, opened 2024-01-19
OffsetTimeSerializer not writing seconds for time object like 15:30:00Z (where seconds is 00)


marceloverdijk:
When having a `OffsetTime` field with e.g. `15:00:00Z` and serializing this with Jackson it outputs:
`15:00Z` and thus the seconds are lost.

Note I have the time module registered via `JavaTimeModule()`.

Although one could argue that the seconds are not needed this gives some problems when validating outputted data against JSON schema.
when using the JSON schema format `time` then the seconds are expected, see https://json-schema.org/understanding-json-schema/reference/string#dates-and-times
without the seconds the JSON schema validator fails with: The value must be a valid time.
(note I'm using [Justify](https://github.com/leadpony/justify) for schema validation)

Looking at the source code, I see this is done on purpose
https://github.com/FasterXML/jackson-modules-java8/blob/master/datetime/src/main/java/tools/jackson/datatype/jsr310/ser/OffsetTimeSerializer.java#L92-L111

```
    private final void _serializeAsArrayContents(OffsetTime value, JsonGenerator g,
            SerializerProvider provider)
        throws JacksonException
    {
        g.writeNumber(value.getHour());
        g.writeNumber(value.getMinute());
        final int secs = value.getSecond();
        final int nanos = value.getNano();
        if ((secs > 0) || (nanos > 0)) {
            g.writeNumber(secs);
            if (nanos > 0) {
                if(useNanoseconds(provider)) {
                    g.writeNumber(nanos);
                } else {
                    g.writeNumber(value.get(ChronoField.MILLI_OF_SECOND));
                }
            }
        }
        g.writeString(value.getOffset().toString());
    }
```

Is it possible to override this behaviour?

Currently I think my only option is create a custom Serializer?

If not possible to override this behaviour would it be at least possible to change this `_serializeAsArrayContents` method from `private` to `protected` so one could simply extend this class and override this method?

marceloverdijk:
Hmm, `JSR310FormattedSerializerBase` (which the `OffsetTimeSerializer` extends) is also package protected. So just making that method `protected` would not help unfortunately.

cowtowncoder:
I would not be against relaxing visibility rules where necessary, fwtw (PRs welcome). But sounds like that might not work.

[...]

PR #386, opened 2026-08-01
Add `JavaTimeFeature.ALWAYS_WRITE_SUBSECOND_DIGITS` (#76)

### LINKED ISSUE #76: Missing milliseconds, when parsing Java 8 date-time, if they are zeros
Issue explained here:
https://stackoverflow.com/questions/47502158/force-milliseconds-when-serializing-instant-to-iso8601-using-jackson

Version: 2.9.5

How to reproduce:
Project where I use Spring Boot 2.0.0.M6, Spring Framework 5.0.1.RELEASE and Jackson 2.9.5

**Test 1: Serialize Instant with milliseconds set to `000`:**

 - Initialize Instant field using `Instant.parse("2017-09-14T04:28:48.000Z")`
 - Serialize it using Jackson
 - Output will be `"2017-09-14T04:28:48Z"`

**Test 2: Serialize Instant with milliseconds set to some non-`000` value:**

 - Initialize Instant field using `Instant.parse("2017-09-14T04:28:48.100Z")`
 - Serialize it using Jackson
 - Output will be `"2017-09-14T04:28:48.100Z"`

**Questions**:

 - Is that behavior by design?
 - Is there anything I can do to force serialization of `000`?

seonwooj0810:
Fixes #76.

### Problem

The JDK ISO formatters omit the sub-second field entirely when it is zero, so the width of a serialized timestamp varies with its value:

```
Instant.parse("2017-09-14T04:28:48.100Z")  ->  "2017-09-14T04:28:48.100Z"
Instant.parse("2017-09-14T04:28:48.000Z")  ->  "2017-09-14T04:28:48Z"       // <-- no ".000"
```

`OffsetDateTime` is worse, since `ISO_OFFSET_DATE_TIME` also strips trailing zeroes *within* the fraction (`.100` is written as `.1`). As reported in the thread, this breaks consumers that expect fixed-precision timestamps, and breaks lexicographic sorting of timestamps stored as text.

### Change

Per @cowtowncoder's [suggestion in the thread](https://github.com/FasterXML/jackson-modules-java8/issues/76#issuecomment-3543564266), the default behaviour is left alone and this adds an opt-in `JavaTimeFeature`:

```java
new JavaTimeModule().enable(JavaTimeFeature.ALWAYS_WRITE_SUBSECOND_DIGITS)
```

When enabled, at least 3 (millisecond) sub-second digits are always written, zero-padded:

| value | default | feature enabled |
|---|---|---|
| `2017-09-14T04:28:48Z` | `2017-09-14T04:28:48Z` | `2017-09-14T04:28:48.000Z` |
| `2017-09-14T04:28:48.100+02:00` | `2017-09-14T04:28:48.1+02:00` | `2017-09-14T04:28:48.100+02:00` |
| `2017-09-14T04:28:48.123456789Z` | `2017-09-14T04:28:48.123456789Z` | `2017-09-14T04:28:48.123456789Z` |

Applies to `Instant`, `OffsetDateTime`, `ZonedDateTime` (both with and without `WRITE_DATES_WITH_ZONE_ID`) and `LocalDateTime`. It affects the **default format only** — an explicit `DateTimeFormatter`, a `@JsonFormat` pattern, and numeric-timestamp serialization are all untouched.

### Notes on the implementation

- It does work with JDK-provided `DateTimeFormatter`, as you hoped — but **not** via `DateTimeFormatterBuilder.appendInstant(3)`, which truncates: it renders `...48.123456789Z` as `...48.123Z`. The formatters in the new `SubSecondFormatters` therefore use `appendFraction(NANO_OF_SECOND, 3, 9, true)`, i.e. minimum 3 digits, maximum 9, so nothing is lost.
- The feature swaps the *default* formatter, not `_formatter`, since a non-null `_formatter` would also force serialization as a JSON String and override `WRITE_DATES_AS_TIMESTAMPS`. That required a small default-format-replacing copy constructor on `InstantSerializerBase` / `LocalDateTimeSerializer`, plus a `withFeatures(JacksonFeatureSet<JavaTimeFeature>)` on each of the four serializers, mirroring how the deserializers already receive module features.
- `LocalTime` / `OffsetTime` are deliberately left out of this first cut: their ISO formats drop the *seconds* field too, which is a somewhat different question (cf. #299). Happy to fold them in if you'd like.
- Naming: you floated `PRESERVE_INSTANT_MSECS_ON_WRITE`; I went with `ALWAYS_WRITE_SUBSECOND_DIGITS` since it also covers the non-`Instant` types and the ">= 3 digits" semantics. Trivial to rename if you prefer yours.

### Verification done

- New test `AlwaysWriteSubSecondDigits76Test` (9 cases). Each case asserts the default output *and* the feature-enabled output side by side, so the previous behaviour is pinned as well as the new. Covers: zero sub-second, higher precision not truncated, all four types, zone-id variant, numeric timestamps unaffected, explicit `@JsonFormat` pattern still wins, and a serialize/parse round-trip (including `+10000-…` and negative years).
- Full build green on the 2.x branch: `./mvnw clean verify` → `Tests run: 1129, Failures: 0, Errors: 0, Skipped: 1`, BUILD SUCCESS.
- Release notes (`VERSION-2.x`, `CREDITS-2.x`) updated; let me know if you'd rather handle those yourself.

Disclosure: this change was prepared with the assistance of an AI coding tool; the design, the `appendInstant` truncation finding and all test cases were reviewed and verified by me against a local build.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -56,8 +56,8 @@ public enum JavaTimeFeature implements JacksonFeature
     /**
      * Feature that determines whether sub-second digits are always written when
      * serializing {@link java.time.Instant}, {@link java.time.OffsetDateTime},
-     * {@link java.time.ZonedDateTime} and {@link java.time.LocalDateTime} as
-     * ISO-8601 Strings using the default format.
+     * {@link java.time.ZonedDateTime} and {@link java.time.LocalDateTime} <b>values</b>
+     * as ISO-8601 Strings using the default format.
      *<p>
      * When disabled (the default), the JDK-provided ISO formatters are used and
      * a zero sub-second value is omitted altogether -- {@code 2017-09-14T04:28:48Z}
@@ -72,6 +72,13 @@ public enum JavaTimeFeature implements JacksonFeature
      * Only affects the default format: an explicit {@code DateTimeFormatter} or
      * a {@link com.fasterxml.jackson.annotation.JsonFormat} pattern takes precedence,
      * as does writing values as numeric timestamps.
+     *<p>
+     * Also note that this only applies to values, and NOT to {@link java.util.Map}
+     * keys: date/time keys keep being written using the JDK-provided ISO formatters,
+     * so a zero sub-second value is still omitted there. Types other than the four
+     * listed above -- notably {@link java.time.LocalTime} and
+     * {@link java.time.OffsetTime}, whose ISO formats also omit the seconds field --
+     * are likewise unaffected.
      *<p>
      * Default setting is disabled, for backwards compatibility.
      *
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -51,7 +51,33 @@ public enum JavaTimeFeature implements JacksonFeature
      *<p>
      * Default setting is false, meaning that Month is serialized/deserialized as a zero-based index.
      */
-    ONE_BASED_MONTHS(false)
+    ONE_BASED_MONTHS(false),
+
+    /**
+     * Feature that determines whether sub-second digits are always written when
+     * serializing {@link java.time.Instant}, {@link java.time.OffsetDateTime},
+     * {@link java.time.ZonedDateTime} and {@link java.time.LocalDateTime} as
+     * ISO-8601 Strings using the default format.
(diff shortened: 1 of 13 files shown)


cowtowncoder:
Since implementation cannot be merged, I'd be +1 for only supporting 3.x for this functionality.

JooHyukKim:
FYI, if going for 3.x support, should directly open PR in Jackson-databind module

seonwooj0810:
Agreed — opened FasterXML/jackson-databind#6151 to add this as `DateTimeFeature.ALWAYS_WRITE_SUBSECOND_DIGITS` on the 3.x side instead. Happy to close this PR whenever, or leave it as a reference for the port — your call.

cowtowncoder:
@seonwooj0810 Since both are done, let's consider them both. Thanks!

PR #388, opened 2026-08-22
Fix #232: add JavaTimeFeature.ALLOW_STRINGIFIED_DURATION_VALUES

### LINKED ISSUE #232: Deserialize Duration from int-like String (e.g. "3600")
I think even after #184 and related changes this is still not possible. 

Trying to parse a string `"3600"` into a Duration field results in an exception: "Text cannot be parsed to a Duration"

I've tried adding `@JsonFormat("SECONDS")` but it seems this only applies to numeric input.

This behaviour is inconsistent with other parts of Jackson:
 - I can convert a int-like String to a regular Long, Integer etc. 
 - I can convert a int-like String to an Instant
 - I cannot convert an int-like String to a Duration
 - 
It would be nice if the handling for  `@JsonFormat` also applied to int-like Strings, not only to actual Numbers

arimu1:
Fixes #232

## Root cause

`DurationDeserializer` parses JSON Strings with `Duration.parse`, which only accepts ISO-8601 (`PT…`). Int-like Strings such as `"3600"` (OAuth `expires_in`) fail even when `@JsonFormat(pattern = "SECONDS")` is present, because that annotation only applies to JSON numbers.

## Change

Add `JavaTimeFeature.ALLOW_STRINGIFIED_DURATION_VALUES` (default **off**, so existing behavior is unchanged). When enabled, integer and decimal numeric Strings are handled the same as JSON numbers:

- integers use `@JsonFormat` unit conversion / `READ_DATE_TIMESTAMPS_AS_NANOSECONDS`
- decimals use seconds with fractional nanos (same as JSON floats)

ISO-8601 duration Strings remain accepted either way.

Enable with:

```java
JsonMapper.builder()
    .addModule(new JavaTimeModule()
        .enable(JavaTimeFeature.ALLOW_STRINGIFIED_DURATION_VALUES))
    .build();
```

## Test plan

- `DurationDeser232Test` (10 tests) — default still fails `"3600"`; feature-on accepts `"3600"` / `"-3600"` / `"60.5"`; `@JsonFormat(pattern="SECONDS")` applies only when enabled
- Neighbors: `DurationDeserTest`, `DurationDeser337Test`, `InstantDeser291Test`
- Full `mvn -pl datetime test` on JDK 21 Temurin: 1143 tests, 0 fail, 1 skip

AI-assisted (Grok 4.6); human-reviewed.

### DIFF
--- a/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
+++ b/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeFeature.java
@@ -84,7 +84,29 @@ public enum JavaTimeFeature implements JacksonFeature
      *
      * @since 2.23
      */
-    ALWAYS_WRITE_SUBSECOND_DIGITS(false)
+    ALWAYS_WRITE_SUBSECOND_DIGITS(false),
+
+    /**
+     * Feature that controls whether stringified numbers (JSON Strings that
+     * without quotes would be legal JSON Numbers) may be deserialized as
+     * {@link java.time.Duration} values (enabled) or not (disabled).
+     * <p>
+     * When disabled (the default), JSON Strings are parsed with
+     * {@link java.time.Duration#parse} and must be ISO-8601 duration
+     * representations such as {@code "PT1H"} -- an int-like String such as
+     * {@code "3600"} fails.
+     * When enabled, integer and decimal numeric Strings are handled the same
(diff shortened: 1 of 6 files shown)


JooHyukKim:
Intention agreed, but 2.x version has been agreed to support mainly bug fixes.

So it is likely that u need to write PR against  Jackson-databind main branch (3.x) repo where our module has been merged into.

PR #390, opened 2026-09-21
Document implicit UTC vs ZonedDateTime serialization #303

### LINKED ISSUE #303: `ZonedDateTime` not serialised as UTC by default altough ObjectMapper Javadocs claiming UTC is the default
As stated in the Jackson ObjectMapper the default should be UTC for writing Json:

    /**
     * Method for overriding default TimeZone to use for formatting.
     * Default value used is UTC (NOT default TimeZone of JVM).
     */

    public ObjectMapper setTimeZone(TimeZone tz) {
        _deserializationConfig = _deserializationConfig.with(tz);
        _serializationConfig = _serializationConfig.with(tz);
        return this;
    }

https://github.com/FasterXML/jackson-databind/blob/b8910125bba81862ea278821ccfd198b1f69fd5d/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L2529-L2537

    /**
     * Base settings contain defaults used for all {@link ObjectMapper}
     * instances.
     */
    protected final static BaseSettings DEFAULT_BASE = new BaseSettings(
            null, // cannot share global ClassIntrospector any more (2.5+)
            DEFAULT_ANNOTATION_INTROSPECTOR,
             null, TypeFactory.defaultInstance(),
            null, StdDateFormat.instance, null,
            Locale.getDefault(),
            null, // to indicate "use Jackson default TimeZone" (UTC since Jackson 2.7)
            Base64Variants.getDefaultVariant(),
            // Only for 2.x; 3.x will use more restrictive default
            LaissezFaireSubTypeValidator.instance,
            // Since 2.12:
            new DefaultAccessorNamingStrategy.Provider(),
            // Since 2.16: [databind#2502] Add a way to configure Caches Jackson uses
            DefaultCacheProvider.defaultInstance()
    );
https://github.com/FasterXML/jackson-databind/blob/b8910125bba81862ea278821ccfd198b1f69fd5d/src/main/java/com/fasterxml/jackson/databind/ObjectMapper.java#L408-L422

With the java 8 time module when the timezone is set to null no time zone information is used on serialisation as the provider.getConfig().hasExplicitTimeZone() does result in false if the default of null is set. So ZonedDateTime is not written with utc as default.

https://github.com/FasterXML/jackson-modules-java8/blob/38f979fbe16a03344f08d6ca899204b7c3bc1afb/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/InstantSerializerBase.java#L147-L163

Is this expected behavior to not overwrite the TimeZone defined in the ZonedDateTime? Than this should be properly Documented.

Minimal example:

        ObjectMapper testmapper = new ObjectMapper();
        testmapper.registerModule(new JavaTimeModule());
        testmapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
        ZonedDateTime testtime = ZonedDateTime.now(ZoneId.of("Europe/Berlin"));
        String teststring = testmapper.writeValueAsString(testtime);
        // 2024-02-20T11:37:42.009045+01:00

        testmapper = new ObjectMapper();
        testmapper.registerModule(new JavaTimeModule());
        testmapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
        testmapper.setTimeZone(TimeZone.getTimeZone("UTC"));
        testtime = ZonedDateTime.now(ZoneId.of("Europe/Berlin"));
        teststring = testmapper.writeValueAsString(testtime);
        // 2024-02-20T10:38:13.0324177Z

AzazelSensei:
ObjectMapper documents UTC as the default TimeZone, but that value stays implicit until setTimeZone is called. ZonedDateTime therefore keeps its own zone when written as ISO-8601 unless WRITE_DATES_WITH_CONTEXT_TIME_ZONE has an explicit timezone.

I documented that in the datetime README, JavaTimeModule, and InstantSerializerBase. Behavior is unchanged.

Fixes #303


cowtowncoder (approved):

cowtowncoder:
Thank you @AzazelSensei !

